PR #3, opened 2024-06-24
Object detection example using tflite interpreter

st-tuanmai:


PaulTR:
So it looks like items that are longer (horizontally) don't mark correctly when scaling.

image1

I also ran into issues when rotating the phone and then rotating back where it didn't fill in the view area again.
image0 (1)

examples/object_detection/ios/RunScripts/download_models.sh
```diff
@@ -0,0 +1,91 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+set -ex
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+MOBILENETV1_SSD_URL="https://storage.googleapis.com/download.tensorflow.org/models/tflite/task_library/object_detection/ios/lite-model_ssd_mobilenet_v1_1_metadata_2.tflite"
```
PaulTR, line 20: Use these:
https://storage.googleapis.com/ai-edge/interpreter-samples/object_detection/ios/efficientdet_lite0.tflite
https://storage.googleapis.com/ai-edge/interpreter-samples/object_detection/ios/efficientdet_lite2.tflite

st-tuanmai: We don't use MOBILENETV1_SSD and EFFICIENTDET_LITE2 right?

PaulTR:
Right well only use two models to keep it simple.

examples/object_detection/ios/RunScripts/download_models.sh
st-tuanmai: @PaulTR 
Their models return 76824 boxes and we need to calculate IOU to remove boxes overlap, then the task will be a little late, sorry about that

st-tuanmai: @PaulTR 
I make a demo in colab using new model, but I don't know how to map box output correctly. (origin, scale).
Please help me check.
Thanks
https://colab.research.google.com/drive/10ShNFLsSGWZPMvFng4HPzDuC3lStcGXH?usp=sharing

examples/object_detection/ios/ObjectDetection/Extensions/CVPixelBufferExtension.swift
```diff
@@ -0,0 +1,157 @@
...
+
+      let srcBytesPerRow = CVPixelBufferGetBytesPerRow(self)
+      let dstBytesPerRow = CVPixelBufferGetBytesPerRow(dstPixelBuffer)
+
+      // Initialize destination buffer with black color
+      memset(dstData, 0, dstBytesPerRow * outputSize)
+
+      // Calculate the scaling factor to fit the source image within the destination square
+      let scaleFactor: Float
+      if srcWidth > srcHeight {
```
priankakariatyml, line 125: replace if with 
``` 
let maxDimension = Float(max(srcWidth, srcHeight))
let scaleFactor = Float(outputSize) / maxDimension
```

examples/object_detection/ios/RunScripts/download_models.sh
PaulTR: If we're not using all of these models, then we should remove the unused ones.

PaulTR:
Since rotation still messes up the viewfinder scaling, just disable rotation in the app for now.

PaulTR:
If this only works with one model, then get rid of the model selection option

st-tuanmai:
> If this only works with one model, then get rid of the model selection option

Hi Paul, this example works correctly on 4 models (EfficientNet Lite 0, EfficientNet Lite 1, EfficientNet Lite 2, SSDMobilenet V1)

st-tuanmai:
> Since rotation still messes up the viewfinder scaling, just disable rotation in the app for now.

I checked and the bounding box is correct when portrait and landscape.

examples/object_detection/ios/ObjectDetection/Views/OverlayView.swift
```diff
@@ -0,0 +1,235 @@
...
+      let yOffset = (viewSize.height - scaledSize.height) / 2
+
+      return (xOffset, yOffset, scaleFactor)
+    }
+
+  /** Calls methods to update overlay view with detected bounding boxes and class names.
+   */
+  func draw(objectOverlays: [ObjectOverlay]) {
+
+    self.objectOverlays = objectOverlays
```
priankakariatyml, line 231: @st-tuanmai You should clear the object overlays here. Sometimes the bounding boxes of the previous frame are visible. 

You can check the MediaPipe code for clearing the bounding boxes. Just add this function. Some variables are not defined in this overlay view. So you can remove those and just retain the lines for clearing the bounding boxes.
https://github.com/google-ai-edge/mediapipe-samples/blob/main/examples/object_detection/ios/ObjectDetector/Views/OverlayView.swift#L86

And then you can call it from this method. You can also check this method needs to be called from somewhere else. Like refer to the MediaPipe sample.

examples/object_detection/ios/ObjectDetection/Views/OverlayView.swift
```diff
@@ -0,0 +1,235 @@
...
+      let xOffset = (viewSize.width - scaledSize.width) / 2
+      let yOffset = (viewSize.height - scaledSize.height) / 2
+
+      return (xOffset, yOffset, scaleFactor)
+    }
+
+  /** Calls methods to update overlay view with detected bounding boxes and class names.
+   */
+  func draw(objectOverlays: [ObjectOverlay]) {
+
```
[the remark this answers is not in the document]
st-tuanmai: @priankakariatyml 
This function is called in the drawAfterPerformingCalculations function, I implement it in lines 99 - 100

PR #10, opened 2024-08-21
Feature implement sample using reinforcement learning tflite

st-tuanmai:


PaulTR:
@st-tuanmai This is an object detection sample, not reinforcement learning? Please submit the correct app.

PR #12, opened 2024-08-30
Android - Text classification - Rename & remove logo

st-haingo:


st-haingo:
Replace by https://github.com/PaulTR/interpreter-samples/pull/11

PR #23, opened 2024-09-05
Update icon and colors digit classification sample

st-tuanmai:


st-tuanmai:
Screenshot 2024-10-11 at 10 29 11
Hi Paul, 
The result always returns a confidence score of 1.0. So I think there is no need to add it anymore.

PR #26, opened 2024-09-05
Update color, app icon super resolution

st-tuanmai:


PaulTR:
This one could use the logo square next to the title.

PR #27, opened 2024-09-05
Update color, app icon text classification

st-tuanmai:


PaulTR:
Let's switch the logo in the toolbar here to just the square version if it's going to be next to the title "Text Classification". The small long logo looks out of place.

PR #43, opened 2024-10-31
[IOS]  image segmenter new model

st-tuanmai:


examples/image_segmentation_new_model/iOS/ImageSegmentation/Configurations/DefaultConstants.swift
```diff
@@ -0,0 +1,68 @@
...
+enum Model: Int, CaseIterable {
+  case isnetPt2eDrq
+  case isnetTflDrq
+  case isnet
+
+
+  var name: String {
+    switch self {
+    case .isnetPt2eDrq:
+      return "Isnet pt2e drq"
```
PaulTR, line 34: Let's only include one model here rather than changing models. Since the original model is ~140mb and then we're quantizing it to make it usable, we only need to include the quantized model.

PaulTR: We can use the isnettfldrq model.

st-tuanmai:
@PaulTR Please help me check again

PaulTR:
From our email, can you update the path from image_segmentation_new_model to image_segmentation_DIS

PR #224, opened 2026-07-09
Add text-to-speech sample (VibeVoice-Realtime-0.5B, streaming AR-diffusion TTS)

john-rocky:
## VibeVoice-Realtime-0.5B TTS (streaming AR next-token-diffusion) on LiteRT

Adds `compiled_model_api/text_to_speech_vibevoice/` — an on-device port of
[VibeVoice-Realtime-0.5B](https://huggingface.co/microsoft/VibeVoice-Realtime-0.5B) (Microsoft, MIT),
a **streaming, autoregressive next-token-diffusion** TTS. First streaming AR-diffusion TTS and first
real-attention autoregressive decoder with an on-device host-side KV cache in these samples.

The 24-layer Qwen2.5-0.5B backbone is split into a 4-layer text LM and a 20-layer TTS LM; each token,
the TTS LM's hidden state conditions a 4-layer DDPM head that a 5-step DPM-Solver++ loop denoises into
a 64-d acoustic latent, which a convolutional σ-VAE decoder turns into 24 kHz audio. **No FFT anywhere.**
The two LMs keep their KV cache host-side as a packed `[1, L·nkv, Pmax, 64]` tensor fed in/out each
step (the ML-Drift-safe "state as graph I/O" pattern).

### Device placement (verified on Pixel 8a, not just desktop parity)

| graph | runs on | why |
|---|---|---|
| base / TTS LM | CPU (fp32) | Mali rejects the KV-step `FULLY_CONNECTED` shape; fp16 collapses the 20-layer stack on ARM XNNPACK |
| diffusion head | GPU (fp32 precision) | compiles + computes correctly |
| σ-VAE decoder | CPU (fp32) | compiles on GPU but ML Drift miscomputes the assembled ConvNeXt block (a graph-assembly buffer bug: every op is bit-exact in isolation) |

Each graph is bit-exact with the PyTorch reference on its chosen accelerator. This is a
correctness-first placement, not a real-time one (a short sentence takes ~30 s on a Pixel 8a — the
two fp32 LMs and the fp32 decoder are the cost).

### Contents
- `conversion/` — build the four graphs + assets from the reference checkpoint (litert-torch).
- `kotlin_cpu_gpu/android/` — Compose app: type text, synthesize on-device, AudioTrack playback.
- Model + card: `mlboydaisuke/VibeVoice-Realtime-0.5B-LiteRT` on Hugging Face.

Note: the Android UI is a Compose wrapper around the same host-loop logic as the device-verified
reference module; the synthesis logic (LMs, DPM-Solver++, connector, decoder) is unchanged.

---

### Correction (2026-07-10)

The placement note above said Mali rejects the KV-step `FULLY_CONNECTED` shape. **That is withdrawn.** The rejection is real on LiteRT **2.1.3** (`INVALID_ARGUMENT: Unsupported weights shape`, `fully_connected.cc:1070`) and **fixed in 2.1.5** — I had observed it once, on 2.1.3, and never re-tested after the runtime moved on. Re-probed on a Pixel 8a with the same graphs, varying only the version:

| graph | litert 2.1.3 | litert 2.1.5 |
|---|---|---|
| `vv_base_lm_kv_fp32` (313 nodes) | compile fail | **313/313 delegated, 27 ms/step** |
| `vv_base_lm_kv_fp16` (341 nodes) | compile fail | **341/341 delegated** |
| `vv_tts_lm_kv_fp32` (1559 nodes, 20L) | — | **1559/1559 delegated, 65 ms/step** |

Numerics too, GPU vs a desktop CPU fp32 reference: base LM hidden corr **0.999964** / `new_k` 1.000000 / `new_v` 0.999988; the 20-layer TTS LM **0.999852 / 0.999998 / 0.999959**.

The sample as filed still runs its LMs on CPU as fp32: it pins 2.1.3, and fp16 independently collapses the 20-layer stack on ARM XNNPACK. Moving them to the GPU is a follow-up (2.1.5 also drops `RELU_0_TO_1`, so the pin is not a free upgrade for every module).

**The σ-VAE decoder finding is unaffected and still open.** Re-verified against the full decoder graph on both runtimes: all 1578 nodes delegate on each, and the GPU output is *bit-identically wrong on both* (std 0.030584, absmax 0.134888, corr 0.0254 vs a CPU fp32 reference whose std is ~0). Deterministic and version-independent.

Docs and comments updated on the branch accordingly.

> **Timing correction (2026-07-10).** The per-step figures above are `run()` **plus the output readback**. `CompiledModel.run()` only enqueues the GPU work — the compute is paid inside the first `readFloat()`. An earlier version of this description quoted the enqueue times (3 ms and 23 ms, written as "10 ms/step" and "48 ms/step"). Delegation and correctness are unchanged.


john-rocky:
Correction pushed: the KV-step `FULLY_CONNECTED` rejection I cited as the reason for CPU placement is a **LiteRT 2.1.3 delegate bug, fixed in 2.1.5**. Same graphs, only the version varied — on 2.1.5 the 4-layer base LM delegates 313/313 nodes (10 ms/step, hidden corr 0.999964 vs CPU) and the 20-layer TTS LM 1559/1559 (48 ms/step, corr 0.999852). A KV-cache autoregressive attention decoder does run on ML Drift.

The samples still ship their LMs on CPU as fp32 — they pin 2.1.3, and fp16 separately collapses these deep stacks on ARM XNNPACK — but the *reason* in the docs and comments was wrong, so I have corrected it on the branch and in the PR description. Moving them to the GPU is a follow-up.

The σ-VAE ConvNeXt miscompute (#224) is unaffected and still open: bit-identically wrong on 2.1.3 and 2.1.5, all 1578 nodes delegated on both.

compiled_model_api/text_to_speech_vibevoice/conversion/README.md
```diff
@@ -0,0 +1,88 @@
...
+to CPU because ML Drift **miscomputes** it at runtime. On-device probing with single-output
+sub-graphs shows every individual op of a ConvNeXt block — conv, RMSNorm, depthwise conv, FFN
+`linear → tanh-GELU → linear` — is bit-exact on the GPU, yet the assembled block miscomputes. The
+divergence first appears at the LayerScale broadcast-multiply that closes the block (the same
+multiply is correct earlier in the block), reproduces on OpenCL and OpenGL and under every
+buffer-storage / precision / constant-sharing option, and persists at fp32 — so it is a
+graph-assembly buffer/scheduling bug, not precision, a backend, or a kernel. Splitting the decoder
+does not help (a single block already trips it), so the decoder runs on CPU, where it is bit-exact
+with the reference. The two LMs also run on CPU, because this sample pins LiteRT 2.1.3, whose Mali
+delegate rejects their KV-step `FULLY_CONNECTED` weights shape — a delegate bug **fixed in 2.1.5**,
```
smilingday: it should be updated in 2.16 already, and whether this is still necessary?

compiled_model_api/text_to_speech_vibevoice/kotlin_cpu_gpu/android/app/src/main/java/com/google/ai/edge/examples/text_to_speech_vibevoice/VibeVoiceSynthesizer.kt
smilingday, line 1: Among Kotlin code, this class is most important and we should try to make it general, without depending on the UI codes and interface is simple, for example, just depends on some model path, and other parameters. Other developers could easily copy some of the codes, and reuse them, and they will have their own UI.

For example, you don't need to pass context here, instead, people just pass the model directory.

Later, we can get some common abstractions for different models.

compiled_model_api/text_to_speech_vibevoice/kotlin_cpu_gpu/android/app/src/main/java/com/google/ai/edge/examples/text_to_speech_vibevoice/VibeVoiceSynthesizer.kt
```diff
@@ -0,0 +1,456 @@
...
+ * Four LiteRT graphs plus host-side orchestration mirror the reference `generate()` —
+ * a stochastic, autoregressive, next-token-diffusion TTS:
+ *
+ *   text --BPE--> ids
+ *   per text token:  embed(host) -> base_lm(CPU) -> +type -> tts_lm(CPU) -> cond
+ *   per speech token (6 per window):
+ *     5-step DPM-Solver++:  head(GPU) x2 (cond + null) -> CFG -> v -> latent[64]
+ *     connector(host) + type -> tts_lm(CPU) & neg tts_lm(CPU) -> next cond
+ *     EOS classifier(host) stops generation
+ *   accumulate latents -> decoder(CPU) -> 24 kHz waveform
```
smilingday, line 46: When make this class more general interfaces, you can add doc about how to use it:

For example:
tts = VibeVoiceSynthesizer(context)
tts.synthesize("How are you?")

compiled_model_api/text_to_speech_vibevoice/README.md
```diff
@@ -0,0 +1,141 @@
...
+Graphs are converted with [litert-torch](https://github.com/google-ai-edge/litert) (per-graph
+tflite-vs-torch corr **1.000000**); the Kotlin host loop and DPM-Solver++ port match the reference
+end-to-end. See [`conversion/`](conversion/).
+
+### Correction (2026-07-10): the LMs are not blocked by the GPU
+
+An earlier version of this sample said ML Drift rejects the KV-step `FULLY_CONNECTED` weights shape,
+so autoregressive attention decoders cannot run on it. **That is withdrawn.** The rejection is real
+on LiteRT **2.1.3** (`INVALID_ARGUMENT: Unsupported weights shape`, `fully_connected.cc:1070`) and
+**fixed in 2.1.5**, where the same graphs delegate fully and compute correctly: `vv_base_lm_kv_fp32`
```
smilingday: update to 2.1.5

compiled_model_api/text_to_speech_vibevoice/README.md
```diff
@@ -0,0 +1,141 @@
...
+**313/313 nodes at 27 ms/step** (hidden corr 0.999964 vs CPU), the 20-layer `vv_tts_lm_kv_fp32`
+**1559/1559 nodes at 65 ms/step** (corr 0.999852). Both timings are `run()` plus the output readback:
+`CompiledModel.run()` only *enqueues*, so timing it alone reports a misleading 3 ms and 23 ms.
+Moving the LMs onto the GPU is a follow-up; this
+sample still pins 2.1.3, and fp16 independently collapses the 20-layer stack on ARM XNNPACK, which is
+a fact about the *CPU* path and remains true.
+
+### The σ-VAE decoder is a confirmed ML Drift correctness bug
```
smilingday, line 59: is this still an issue?
Then file a issue to LiteRT repo?

compiled_model_api/text_to_speech_vibevoice/README.md
```diff
@@ -0,0 +1,141 @@
...
+| :-- | :-- | :--: | :-- |
+| `vv_base_lm_kv_fp32` | x[1,1,896], cos/sin[1,1,1,64], mask[1,1,1,129], pk/pv[1,8,128,64] → hidden[1,1,896], k/v[1,8,1,64] | **CPU (fp32)** | This sample pins LiteRT 2.1.3, whose Mali delegate rejects the KV-step `FULLY_CONNECTED` weights shape — **fixed in 2.1.5**, see below. fp16 separately collapses the stack on ARM XNNPACK. |
+| `vv_tts_lm_kv_fp32` | same I/O, `Pmax=384`, `L*nkv=40` | **CPU (fp32)** | Same as above (20 layers). |
+| `vv_diffhead_fp16` | noisy[1,64], t_freq[1,256], cond[1,896] → v[1,64] | **GPU (fp32 precision)** | Small; compiles and computes correctly on ML Drift. |
+| `vv_decoder_fp32` | latent[1,64,128] → wav[1,1,409600] | **CPU (fp32)** | Compiles on GPU but ML Drift **miscomputes** it — see below. |
+
+This is a **hybrid** placement, not an all-GPU one: only the tiny diffusion head runs on the GPU.
+Graphs are converted with [litert-torch](https://github.com/google-ai-edge/litert) (per-graph
+tflite-vs-torch corr **1.000000**); the Kotlin host loop and DPM-Solver++ port match the reference
+end-to-end. See [`conversion/`](conversion/).
```
smilingday, line 36: We cam think about our samples could serve different needs, and more modulized and different users or agents could pick what they want most.

- Users:  developers or agents to follow;
- Models: could download from HuggingFace, or convert themselves;
- Purpose: could be just demos, how to integrate (those VoiceSynthethizer Kotlin API is most important, and people can copy), or conversion scripts (learn how to convert model, or agents learn those tricks).

smilingday (commented):
Could you put all text_to_speech models under same folder, to make it easy to organize?

compiled_model_api/text_to_speech_vibevoice/conversion/README.md
john-rocky: You're right — the CPU placement of the two LMs was no longer necessary, and this paragraph was stale (the sample has pinned LiteRT 2.1.5 all along; the "pins 2.1.3" wording was a leftover from an earlier revision). Updated in 13bb4e7d:

- **Both Qwen2 LMs now run on the GPU** (`GpuOptions` `Precision.FP32`). The `FULLY_CONNECTED` rejection was a 2.1.3 delegate bug, fixed in 2.1.5 (and 2.1.6). Re-verified this sample on a Pixel 8a: all three GPU graphs fully delegate (313/313, 1559/1559, 167/167 with `LITERT_CL`), and a 3.2 s utterance now generates in 34.2 s, vs ~72 s with the LMs on CPU. With FP32 precision the end-to-end waveform is bit-identical to the CPU reference (corr 1.000000; the fp16 default measures 0.991886), at a 3.9% time cost.
- **The σ-VAE decoder still needs CPU.** Its miscompute is a separate graph-assembly bug, unrelated to the `FULLY_CONNECTED` rejection, and reproduces bit-identically on 2.1.3 and 2.1.5 (all 1578 nodes delegate; corr 0.0254 vs reference). I haven't re-probed it on 2.1.6 yet — if it computes correctly there, the decoder can move too in a follow-up.
- **Why the pin stays 2.1.5 rather than 2.1.6:** the LM KV-step graphs are multi-output (`hidden`, `new_k`, `new_v`), which is the shape affected by the output-aliasing issue (google-ai-edge/LiteRT#8599) whose affected output differs between 2.1.5 and 2.1.6, and 2.1.5 is the configuration the end-to-end A/B was validated on. The 2.1.6 broadcast fix (google-ai-edge/LiteRT#8593) doesn't affect anything this model uses.

john-rocky: Follow-up: re-probed the σ-VAE decoder on **2.1.6** (native `CompiledModel` harness, Pixel 8a, same seeded latent input, GPU-only so full residency is required):

| Run | std | absmax | corr vs CPU |
|---|---|---|---|
| CPU (XNNPACK, reference) | ~0 | 6.5e-5 | — |
| GPU 2.1.6 (1578/1578 delegated) | 0.0432 | 0.1927 | **0.0146** |
| GPU 2.1.6, `Precision.FP32` | 0.0449 | 0.2006 | 0.0150 |
| GPU ≤2.1.5 harness (control) | 0.0432 | 0.1927 | 0.0146 |

Still miscomputes on 2.1.6, and the 2.1.6 GPU output is **byte-identical** to the ≤2.1.5 wrong output on the same input (`np.array_equal` → true), so none of the 2.1.6 fixes touch it. The decoder stays on CPU; I've kept the README's placement rationale as is.

PR #226, opened 2026-07-09
Add Dia2-1B two-speaker dialogue TTS sample (RQ-Transformer on CompiledModel)

john-rocky:
Adds `compiled_model_api/text_to_speech_dia2`: [Dia2-1B](https://huggingface.co/nari-labs/Dia2-1B) (Nari Labs, Apache-2.0), a two-speaker **dialogue** TTS model, running fully on-device on LiteRT `CompiledModel`. Type a `[S1]`/`[S2]` script and the app speaks it.

Dia2 is a Moshi-style **RQ-Transformer**: once per 12.5 Hz frame a 30-layer *temporal* transformer emits a word-timing action plus [Mimi](https://huggingface.co/kyutai/mimi) codebook 0, and a 3-layer *depformer* then autoregressively fills the remaining 31 codebooks for that same frame. Mimi's 32 quantizers decode to 24 kHz audio.

## Device placement

Every graph runs on the **CPU as fp32**. The GPU delegate rejects the language models' KV-step `FULLY_CONNECTED` weight shapes, and fp16 collapses these deep stacks on ARM. The KV caches, RoPE, the 34-channel embedding sum, the depformer's projections and all sampling live in Kotlin, so the graphs are pure step functions: each takes the packed cache as an input and returns the new key/value slice, which the host appends at the tail — no `SCATTER`/`GATHER`, no dynamic shapes.

## Three things that are easy to get wrong

1. **Both text streams carry real word tokens**, not new-word/pad markers. On a new word the main stream emits the word's first text token while the second stream emits `NEW_WORD`; during the padding frames that follow, the main stream drains the rest of the word and the second stream drains a two-word lookahead. Feeding markers instead produces fluent, confident, completely wrong speech.
2. **Undo the delay pattern before decoding** (16 frames for cb0, 18 for the rest), or the audio is muffled.
3. **Mimi decode must be one pass.** Its decode path is upsample → *causal* decoder transformer → SEANet, so the receptive field is unbounded. Decoding disjoint windows costs ~13% relative error (corr 0.991 vs a full-sequence PyTorch decode); one 256-frame window with a zeroed tail is exact — corr **0.999999**.

## The speaker is sampled, so the voice prompt is baked in

With no voice prefix Dia2 **samples the speaker identity** and the voice changes every run (median F0 wanders ~120 Hz). Classifier-free guidance does **not** pin it — over 8 matched seeds the F0 spread is 144 Hz at `cfg_scale=1.0` and 134 Hz at `2.0`; guidance only steadies levels. The model's remedy is a voice prefix, which normally needs Whisper word timings and a Mimi *encoder* (both host-only). `conversion/bake_prefix.py` precomputes the prompt into a 13 kB JSON that ships in `assets/`, so the device only replays the warm-up and **no Mimi encoder ships**. On device the speakers then track their prompts (S1 214 Hz / S2 114 Hz against prompts of 247 Hz / 88 Hz; without a prefix S2 never drops below 214 Hz).

CFG (`cfg_scale = 2.0`, Dia2's default) is implemented faithfully and is subtle: the guided logits `uncond + scale·(cond − uncond)` only *select* the top-k candidate set, while the draw is a temperature softmax over the **conditional** logits restricted to that set. That needs a second unconditional branch with its own KV cache.

## Validation

Every ported component was checked against the reference PyTorch implementation on host before it reached the device:

| Component | Check | Result |
|---|---|---|
| Word-pacing state machine | vs recorded reference frame stream | 0/60 mismatches |
| Tokenizer + script parsing | vs reference entries | 10/10 exact |
| Depformer 31-stage KV glue | vs PyTorch depformer | corr 1.0000, argmax 31/31 |
| Mimi decode | vs PyTorch full-sequence decode | corr 0.999999 |
| Voice-prefix warm-up | vs reference warm-up + generation | 0 mismatches (71 + 63 frames) |

Device-verified on a Pixel 8a: 4.0 s of audio in 190 s (correctness-first, not real-time), peak ~4.6 GB RSS.

Weights: https://huggingface.co/mlboydaisuke/Dia2-1B-LiteRT

---

### Correction (2026-07-10)

The placement note above said the GPU delegate rejects the language models KV-step `FULLY_CONNECTED` weight shapes. **That is withdrawn.** The rejection is real on LiteRT **2.1.3** and **fixed in 2.1.5**, where a 20-layer Qwen2 KV-step decoder delegates 1559/1559 nodes at 65 ms/step and matches CPU to corr 0.999852.

**Dia2 was not delegate-blocked either — both blockers were in my own graph.**

1. *Compile.* Only 31 of 249 nodes delegated. logcat: `RESHAPE: ... has bad input dims size: 5` — my fused-QKV authoring did `view(1, 1, 3, NH, HD)`, rank 5, above ML Drift maximum rank of 4. Slicing the last dimension into thirds gives **237/237 nodes**.
2. *Correctness.* It then compiled but computed wrong (corr 0.9604), and `GpuOptions.Precision.FP32` did **not** change that (0.960297) — a miscompute, not fp16 accumulation. Cause: the known BMM + broadcast-`ADD` bug, `softmax(q@kᵀ + mask)` with the mask broadcasting over the head dimension. Pre-expanding it host-side from `[1,1,1,D]` to `[1,NH,1,D]` gives corr **0.999975 / 0.999992 / 0.999989** at default precision and **1.000000** at FP32.

The sample as filed still runs everything on CPU as fp32, because fp16 genuinely does collapse these deep stacks on ARM. Moving the depformer onto the GPU is a follow-up; the 3.0 GB temporal graph has not been evaluated there yet.

Docs and comments updated on the branch accordingly.

---

### Follow-up (2026-07-10, device-measured): the depformer runs on the GPU, and it is *not worth it*

Having established that the delegate was never the blocker, I actually moved the depformer onto the GPU and measured it end-to-end on a Pixel 8a.

**Correctness: perfect.** All three weight sets delegate **237/237 nodes**, and the generated audio is bit-identical to the CPU path — **4288/4288 codebook tokens equal**, waveform correlation **1.000000**, RMS difference **0.000000**, over a 133-frame utterance with a fixed sampling seed.

**Speed: no win.** Per-graph timing, same seed, 3906 depformer calls:

| | KV upload | small inputs | `run()` | output readback | depformer total |
|---|---|---|---|---|---|
| CPU | 0.45 s | 0.28 s | 75.8 s | 0.58 s | **76.4 s** |
| GPU | 2.0 s | 2.2 s | 17.1 s | 61.2 s | **78.3 s** |

`CompiledModel.run()` only **enqueues**; the GPU work is paid inside the first `readFloat()`. Timing `run()` alone suggests a 4.4× win that does not exist. Per call the GPU costs **21.1 ms** against the CPU's **19.7 ms** — a 3-layer, single-token step graph is too small to amortise dispatch and synchronisation.

So the sample keeps everything on CPU, and the stated reason changes from "the delegate rejects it" (false) to "the GPU is correct but offers no gain at this graph size". The 30-layer temporal transformer is the graph worth moving, but at 3.0 GB fp32 it does not fit in GPU memory as exported.

> **Timing correction (2026-07-10).** The per-step figures above are `run()` **plus the output readback**. `CompiledModel.run()` only enqueues the GPU work — the compute is paid inside the first `readFloat()`. An earlier version of this description quoted the enqueue times (3 ms and 23 ms, written as "10 ms/step" and "48 ms/step"). Delegation and correctness are unchanged.


john-rocky:
Correction pushed: the KV-step `FULLY_CONNECTED` rejection I cited as the reason for CPU placement is a **LiteRT 2.1.3 delegate bug, fixed in 2.1.5**. Same graphs, only the version varied — on 2.1.5 the 4-layer base LM delegates 313/313 nodes (10 ms/step, hidden corr 0.999964 vs CPU) and the 20-layer TTS LM 1559/1559 (48 ms/step, corr 0.999852). A KV-cache autoregressive attention decoder does run on ML Drift.

The samples still ship their LMs on CPU as fp32 — they pin 2.1.3, and fp16 separately collapses these deep stacks on ARM XNNPACK — but the *reason* in the docs and comments was wrong, so I have corrected it on the branch and in the PR description. Moving them to the GPU is a follow-up.

The σ-VAE ConvNeXt miscompute (#224) is unaffected and still open: bit-identically wrong on 2.1.3 and 2.1.5, all 1578 nodes delegated on both.

john-rocky:
Follow-up, device-measured: I moved the depformer onto the GPU. It delegates **237/237 nodes** per weight set and produces **bit-identical audio** (4288/4288 codebook tokens equal, waveform corr 1.000000, RMS diff 0.000000). But it is **not faster** — 3906 depformer calls cost 76.4 s on CPU and 78.3 s on GPU.

The trap: `CompiledModel.run()` only *enqueues*. The GPU work is paid inside the first `readFloat()` (17.1 s enqueue + 61.2 s readback). Timing `run()` alone suggests a 4.4× win that does not exist. Per call the GPU costs 21.1 ms against the CPU's 19.7 ms — a 3-layer, single-token step graph is too small to amortise dispatch and synchronisation.

The sample therefore stays on CPU, with the *reason* corrected. Branch and description updated.

PR #239, opened 2026-07-23
refactor: rename samples/litertlm to samples/litert_lm

smilingday:


yuhuichen1015 (approved):
Thanks a lot for the change!

PR #241, opened 2026-07-24
Add Qwen3-TTS Tensor API implementation (talker step graph, codec decoder, verification suite)

john-rocky:
Thanks for setting up the new `models/` layout (#238) and the
`models/qwen/qwen3_tts/tensor_api/` home — this PR fills that directory
with the Qwen3-TTS Tensor API reference implementation. No weights are
included, and nothing here is time-sensitive; happy to reshape any of it
to your conventions.

**What this adds**

- `talker_step/` — the talker decode-step graph authored with the C++
  Tensor API: `prefill`/`decode` signatures sharing weights, fixed-size KV
  caches as signature I/O (DynamicUpdateSlice at a runtime position,
  zero-copy output→input rebind), codebook-0 pick + 15-group code
  predictor + next-frame feedback aggregation in-graph — one signature
  call per 12.5 Hz frame. Weight paths: fp32, int8 per-channel, int4
  blockwise-32, pre-quantized (GPTQ) companion files. Host-staged additive
  bias input covers suppression, EOS gating, min-new-tokens, and exact
  temperature sampling (Gumbel noise-as-input).
- `codec_decode/` — the full speech_tokenizer decoder (RVQ + transformer
  + vocoder) as one Tensor API graph.
- `verify/` — independent numpy references for both graphs (the C++
  graphs are bit-exact against them on the real checkpoint), the GPTQ
  quantizer emitting the exact blockwise-32 layout the loader consumes,
  speaker x-vector enrollment, and an end-to-end text→wav smoke driver.
- `talker_step/attn_bench.cc` — attention micro-benchmark: raw ops vs
  `odml.scaled_dot_product_attention` vs the `odml.runtime_bmm` +
  `odml.cache_update` composite pair, with CPU decompositions as the
  numeric reference; includes multi-step feedback-loop runs (LiteRT
  PR #8796).
- The models adopt `odml.runtime_bmm` and `odml.rms_norm` via build flags.

**Build**

BUILD files target this repository's `@litert_archive` workspace; the one
helper whose `select()` keys don't survive a cross-repository load
(`commandline_flag_copts`) is inlined. One dependency is adapted to
current `main`: the binaries no longer depend on
`litert/cc:litert_api_with_dynamic_runtime` (matching the current gemma3
example) — with it, the runtime gets linked twice on `main` (duplicate
`GetLiteRtRuntimeBuiltin` at the Android link). From the repo root:

```
bazel build --config=macos_arm64 //models/qwen/qwen3_tts/tensor_api/talker_step:talker_main
bazel build --config=macos_arm64 //models/qwen/qwen3_tts/tensor_api/codec_decode:codec_main
ANDROID_NDK_HOME=<ndk> bazel build -c opt --config=android_arm64 \
  --incompatible_enable_cc_toolchain_resolution \
  --incompatible_enable_android_toolchain_resolution \
  //models/qwen/qwen3_tts/tensor_api/talker_step:talker_main
```

The Android arm64 binaries (talker_main, attn_bench, codec_main) build
and link from this workspace against current LiteRT `main`; on macOS all
sources compile, and the final link stops at a workspace-level signing
issue described below.

**Provenance and verification**

- Sources are the same as
  [john-rocky/litert-tensor-audio-examples](https://github.com/john-rocky/litert-tensor-audio-examples),
  adapted to this layout. Validated at LiteRT `a19d8fa` + the PR #8796
  runner change; if anything drifts against current `main`, tell us and
  we will chase it.
- Measured highlights are in the README (Pixel 8a CPU GPTQ int4 99.6
  ms/frame = RTF 1.24; M4 Max Metal GPTQ int4 + composite stack 17.3
  ms/frame = RTF 0.216; two-window measurement protocol).
- We kept the standalone bug-repro tools (the int4 two-signature Metal
  repro, the delegate-init OOM logger) in the external repo to keep this
  sample user-facing — happy to add them here too if you'd rather have
  everything in one place.

**Three build observations on the current workspace (independent of this
PR — sharing them since a reviewer will hit them before reaching this
code)**

- The checked-in `.bazelrc.user` overrides the Apple toolchain config
  with a crosstool target that doesn't exist in the pinned apple_support
  (plus a machine-local Xcode path and an iPhoneSimulator SDKROOT), so
  any `--config=macos*` build fails in analysis until that file is
  removed — it may have been committed unintentionally.
- With it removed, LiteRT `main` compiles from source end-to-end here,
  but the `SignBinary` action for `litert_runtime_c_api` fails: the
  rules_apple `codesigningtool` py stub comes out with unsubstituted
  rules_python placeholders (`%interpreter_args%` …) under the hermetic
  Python setup; same result with rules_apple 4.5.0 and 3.22.0. All our
  translation units compile cleanly against LiteRT `main`; the dylib
  signing step is where a macOS build currently stops.
- `build:android_arm64` in this repository's `.bazelrc` is missing the
  two toolchain-resolution lines LiteRT's own `.bazelrc` has
  (`--incompatible_enable_cc_toolchain_resolution` and
  `--incompatible_enable_android_toolchain_resolution`); since the NDK
  is wired via rules_android_ndk (resolution-style only), plain
  `--config=android_arm64` fails analysis. Passing both flags on the
  command line works — the READMEs note this until the config carries
  them.

Happy to send a small separate PR for any of these if that's welcome —
otherwise just flagging what we saw.

### DIFF
--- a/models/qwen/qwen3_tts/tensor_api/codec_decode/BUILD
+++ b/models/qwen/qwen3_tts/tensor_api/codec_decode/BUILD
@@ -1,0 +1,89 @@
+# Copyright 2026 The Google AI Edge Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load(
+    "@litert_archive//litert/build_common:special_rule.bzl",
+    "litert_android_linkopts",
+    "litert_gpu_accelerator_prebuilts",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "codec_config",
+    hdrs = ["codec_config.h"],
+)
+
+cc_library(
+    name = "codec_graph",
+    srcs = ["codec_graph.cc"],
+    hdrs = ["codec_graph.h"],
+    deps = [
+        ":codec_config",
+        "@litert_archive//tensor",
+        "@litert_archive//tensor:arithmetic",
+        "@litert_archive//tensor:buffer",
+        "@litert_archive//tensor:datatypes",
+        "@litert_archive//tensor/backends/tflite:arithmetic_tflite",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/log:absl_check",
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_library(
+    name = "codec_weights",
+    srcs = ["codec_weights.cc"],
+    hdrs = ["codec_weights.h"],
+    deps = [
+        ":codec_config",
+        ":codec_graph",
+        "@litert_archive//tensor",
+        "@litert_archive//tensor:buffer",
+        "@litert_archive//tensor:datatypes",
+        "@litert_archive//tensor/examples/gemma3:safetensor_loader",
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_binary(
+    name = "codec_main",
+    srcs = ["codec_main.cc"],
+    copts = ["-DGOOGLE_COMMANDLINEFLAGS_FULL_API=1"] + select({
+        "@litert_archive//litert:android": ["-DABSL_FLAGS_STRIP_NAMES=0"],
+        "//conditions:default": [],
+    }),
+    data = litert_gpu_accelerator_prebuilts(),
+    linkopts = litert_android_linkopts(),
+    deps = [
+        ":codec_config",
+        ":codec_graph",
+        ":codec_weights",
+        "@litert_archive//litert/cc:litert_common",
+        "@litert_archive//litert/cc:litert_environment",
+        "@litert_archive//litert/cc:litert_options",
+        "@litert_archive//litert/cc/options:litert_gpu_options",
+        "@litert_archive//tensor",
+        "@litert_archive//tensor:datatypes",
+        "@litert_archive//tensor/backends/tflite:tflite_flatbuffer_conversion",
+        "@litert_archive//tensor/runners/litert:litert_dynamic_runner",
+        "@com_google_absl//absl/flags:flag",
+        "@com_google_absl//absl/flags:parse",
+        "@com_google_absl//absl/status",
+    ],
+)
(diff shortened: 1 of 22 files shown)


models/qwen/qwen3_tts/tensor_api/codec_decode/BUILD
```diff
@@ -0,0 +1,89 @@
...
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_binary(
+    name = "codec_main",
+    srcs = ["codec_main.cc"],
+    copts = ["-DGOOGLE_COMMANDLINEFLAGS_FULL_API=1"] + select({
```
pyu10055, line 67: commandline_flag_copts() from litert_build_defs.bzl is used. If available in the @litert_archive import within litert-samples, preferring commandline_flag_copts() will keep platform-specific flag definitions centralized and consistent with core LiteRT targets.

pyu10055: same for the talker_step BUILD file

pyu10055 (approved):

john-rocky:
Thank you for the review and the approval, and for pointing at `commandline_flag_copts()` — keeping the platform flag definitions centralized is the right instinct, and I wanted to adopt it.

I checked against the `@litert_archive` snapshot this repository pulls: the helper is there, but it isn't consumable from a litert-samples package as currently imported, so I've left the explicit flags in place. Nothing here blocks the merge.

Details, if useful:

- `litert_build_defs.bzl` is reachable and `commandline_flag_copts()` is defined in it, so the symbol itself is available.
- Its `select()` key is the unqualified `//litert:android`. Evaluated from a litert-samples package that resolves against this repository rather than `@litert_archive`, and analysis stops with `no such package 'litert'` / `errors encountered resolving select() keys`. Rewriting just that key to `@litert_archive//litert:android` in a local copy makes the error go away, so the label is the only thing in the way.
- This repository's `WORKSPACE` already carries a `patch_cmds` sed that rewrites `//litert` to `@litert_archive//litert` for `special_rule.bzl` — that is why `litert_android_linkopts()` works from these same files. `litert_build_defs.bzl` is not covered by that patch.
- One behavioural difference as well: the helper scopes `-DGOOGLE_COMMANDLINEFLAGS_FULL_API=1` to Android only, while these targets set it on every platform. The primary documented build path for these binaries is `--config=macos_arm64`, which would lose the define.

Your call on how you'd like this handled on the LiteRT side.

jl45G (approved):

smilingday:
Seems get approvals, and @john-rocky can address the comments, and then we can merge.

models/qwen/qwen3_tts/tensor_api/codec_decode/BUILD
john-rocky: Thanks again for the pointer. Re-verified today against current LiteRT main (which `@litert_archive` tracks): the situation in my earlier comment on the conversation tab (https://github.com/google-ai-edge/litert-samples/pull/241#issuecomment-5086836069) is unchanged — `litert_build_defs.bzl` loads, but `commandline_flag_copts()` still isn't consumable from a litert-samples package, because its `//litert:android` select key resolves against this repository and the WORKSPACE `patch_cmds` rewrite that fixes exactly this for `special_rule.bzl` doesn't cover `litert_build_defs.bzl`. So the explicit flags stay here and in the talker_step BUILD for now.

Happy to adopt the helper across all the copts sites (this file, talker_step, and the one in #242) in a follow-up once it's consumable — extending the same one-line `patch_cmds` rewrite to `litert_build_defs.bzl` looks sufficient, and I'd be glad to send that change. Resolving this thread so it doesn't hold up the merge; happy to reopen if you'd prefer it handled differently.

PR #242, opened 2026-07-24
Add LLaDA-8B diffusion-LM Tensor API implementation (in-graph denoise step)

john-rocky:
Following the invitation to create new model directories: this adds
`models/llada/llada_8b/tensor_api/` — a diffusion-LM denoise step for
LLaDA-8B-Base authored with the C++ Tensor API. Small PR (7 files), no
weights included.

- One in-graph denoise step: bidirectional backbone + ScatterNd-free
  top-k remasking (TopK → OneHot → Sum → SelectV2); the host only feeds
  ids back between steps. `--llada` selects the 8B architecture and the
  sharded-safetensors loader; composite options `--attention=raw|sdpa`
  and `--norms=raw|composite`; in-process int8/int4 weight quantization;
  `--reuse_tflite` for on-device runs without the checkpoint.
- Correctness: identical per-step id trajectories vs an independent numpy
  reference on real weights, full schedule; int4 ids identical to fp32.
  The full 32L model runs on a Pixel 8a CPU (memory-mapped 6.29 GB
  model) with ids identical to desktop.
- Same build story as the Qwen3-TTS tensor_api PR (BUILD targets
  `@litert_archive`, inlined copts helper, one dependency adapted to
  current `main`); runtime behavior validated at LiteRT `a19d8fa` + PR
  #8796, and the Android arm64 binary builds and links from this
  workspace against current `main` — we will chase anything else that
  drifts.
- Kernel-shape note in the README that may interest the XNNPACK side:
  blockwise-int4 fast kernels are GEMV-only today, so full-sequence
  diffusion steps land on a slow path — per-channel int8 is the right
  CPU default for this model class (measured 8L probe: int8 81.6 vs int4
  791.5 ms/step on M4 Max).

### DIFF
--- a/models/llada/llada_8b/tensor_api/BUILD
+++ b/models/llada/llada_8b/tensor_api/BUILD
@@ -1,0 +1,83 @@
+# Copyright 2026 The Google AI Edge Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load(
+    "@litert_archive//litert/build_common:special_rule.bzl",
+    "litert_android_linkopts",
+    "litert_gpu_accelerator_prebuilts",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "diffusion_graph",
+    srcs = ["diffusion_graph.cc"],
+    hdrs = ["diffusion_graph.h"],
+    deps = [
+        "@litert_archive//tensor",
+        "@litert_archive//tensor:arithmetic",
+        "@litert_archive//tensor:buffer",
+        "@litert_archive//tensor:datatypes",
+        "@litert_archive//tensor/backends/tflite:arithmetic_tflite",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/log:absl_check",
+        "@com_google_absl//absl/strings",
+        "@flatbuffers",
+    ],
+)
+
+cc_library(
+    name = "diffusion_weights",
+    srcs = ["diffusion_weights.cc"],
+    hdrs = ["diffusion_weights.h"],
+    deps = [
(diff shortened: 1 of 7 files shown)


models/llada/llada_8b/tensor_api/BUILD
```diff
@@ -0,0 +1,83 @@
...
+        "@com_google_absl//absl/status",
+        "@com_google_absl//absl/status:statusor",
+        "@com_google_absl//absl/strings",
+    ],
+)
+
+cc_binary(
+    name = "diffusion_main",
+    srcs = ["diffusion_main.cc"],
+    copts = ["-DGOOGLE_COMMANDLINEFLAGS_FULL_API=1"] + select({
```
pyu10055, line 61: If commandline_flag_copts() is available via the @litert_archive//litert/build_common:litert_build_defs.bzl import, adopting it will maintain build-flag consistency across targets and platforms.

pyu10055 (approved):

john-rocky:
Thank you for the review and the approval, and for pointing at `commandline_flag_copts()` — keeping the platform flag definitions centralized is the right instinct, and I wanted to adopt it.

I checked against the `@litert_archive` snapshot this repository pulls: the helper is there, but it isn't consumable from a litert-samples package as currently imported, so I've left the explicit flags in place. Nothing here blocks the merge.

Details, if useful:

- `litert_build_defs.bzl` is reachable and `commandline_flag_copts()` is defined in it, so the symbol itself is available.
- Its `select()` key is the unqualified `//litert:android`. Evaluated from a litert-samples package that resolves against this repository rather than `@litert_archive`, and analysis stops with `no such package 'litert'` / `errors encountered resolving select() keys`. Rewriting just that key to `@litert_archive//litert:android` in a local copy makes the error go away, so the label is the only thing in the way.
- This repository's `WORKSPACE` already carries a `patch_cmds` sed that rewrites `//litert` to `@litert_archive//litert` for `special_rule.bzl`, which is why `litert_android_linkopts()` works from this same file. `litert_build_defs.bzl` is not covered by that patch.
- One behavioural difference as well: the helper scopes `-DGOOGLE_COMMANDLINEFLAGS_FULL_API=1` to Android only, while this target sets it on every platform. The primary documented build path for this binary is `--config=macos_arm64`, which would lose the define.

Your call on how you'd like this handled on the LiteRT side.

jl45G (approved):

models/llada/llada_8b/tensor_api/BUILD
john-rocky: Thanks for the pointer here as well. Same result as #241, re-verified today: the helper loads from `@litert_archive`, but `commandline_flag_copts()` still isn't consumable from a litert-samples package (details in my earlier comment on the conversation tab: https://github.com/google-ai-edge/litert-samples/pull/242#issuecomment-5086836355), so the explicit flags stay for now. Happy to switch this site together with the #241 ones in a follow-up that extends the WORKSPACE `patch_cmds` rewrite to `litert_build_defs.bzl`. Resolving this thread so it doesn't hold up the merge — glad to reopen if you'd prefer a different route.

PR #244, opened 2026-07-29
Add text-to-image diffusion sample (Bonsai Image 4B)

john-rocky:
A full diffusion text-to-image pipeline running on LiteRT: three fixed-shape tflite graphs (text encoder, DiT, VAE decoder) plus a ~150-line host loop with no torch and no diffusers at inference time — same shape as the text_to_speech_lm sample's host-loop approach.

The model is PrismML Bonsai Image 4B, a ternary-weight DiT on the FLUX.2-klein-4B architecture. Its ternary weights land losslessly as {-7, 0, +7} in the int4 block-32 container, so the 2.11 GiB DiT file is a lossless container for the released weights. Converted models are published at https://huggingface.co/litert-community/Bonsai-Image-ternary-4B (Apache-2.0 upstream).

Included:
- `python/generate.py` — host loop: tokenize → FlowMatch-Euler over the DiT → latent unpatchify → VAE decode (deps: ai-edge-litert, numpy, Pillow, transformers)
- `conversion/` — complete reproduction recipe, including the two conversion walls this model hits (Flux2's float64 RoPE frequency table vs `tfl.pow` legalization, and zero min-max scales on all-zero ternary blocks that XNNPACK rejects)

Measured: ~19 s per 512×512 image (4 steps) on a MacBook CPU; ~64 s on iPhone 17 Pro (XNNPACK, 6 threads, 2.9 GiB peak), with device output bit-exact against the desktop run. One note carried into the README: through the C API the XNNPACK delegate must be attached explicitly — the reference-kernel fallback is both much slower and numerically different on blockwise-int4 weights.

Happy to adjust placement, naming, or scope to whatever fits the repo best.


john-rocky:
Restructured to the current `models/` layout — `models/bonsai/bonsai_image_4b/` with the host loop under `python/` and the conversion recipe under `converted/`, matching `models/qwen/qwen3_tts` and `models/gemma`. Content is unchanged. Happy to rename the family/surface directories if a different convention is preferred.

smilingday (approved):

PR #245, opened 2026-07-29
Add litert_gpu_toolkit to utilities/

john-rocky:
`utilities/README.md` lists "Common conversion utilities" as one of the things this directory is for. This adds a set of them.

These are the pre-conversion patches that rewrite common PyTorch patterns into forms the GPU delegate accepts, plus a checker that reads the flatbuffer and reports GPU-incompatible ops, delegate-only ops, Flex ops and tensors above rank 4. They came out of converting real models and hitting walls on device, and were re-implemented inline across roughly 40 conversion scripts until now.

Each patch is checked numerically against its PyTorch reference and is device-proven in at least one shipped sample. The README maps every patch to the pattern it rewrites and is explicit about the limits rather than leaving them to be discovered:

- exact: the zero-stuff ConvTranspose rewrite, and the tent-matmul `grid_sample` (~2e-7 including zeros-padding out of bounds)
- approximations: the GELU and Swish substitutions, where the approximation choice can matter for wide-output-range heads
- `patch_safe_layernorm`'s default mode reduces in a scaled domain and is therefore not bit-faithful to stock LayerNorm; `scale="adaptive"` is the bit-faithful mode
- `hierarchical_mean` is exact only for power-of-two spatial dims

Nothing here is wired into a build, so merging changes no existing sample. `utilities/README.md` itself is unmodified. Happy to split this, rename it, or move it if another location fits the layout better.

### DIFF
--- a/utilities/litert_gpu_toolkit/checker.py
+++ b/utilities/litert_gpu_toolkit/checker.py
@@ -15,45 +15,18 @@
 """
 Post-conversion GPU verification for TFLite models via LiteRT CompiledModel.
 
-The verdict comes from the LiteRT CompiledModel API itself: the model is
-compiled for the GPU accelerator, every signature is run on random inputs,
-and the outputs are compared against a CPU-compiled reference. The static
-op lists below are advisory only — they point at the patch to apply when
-GPU compilation fails, but they are not the source of truth and may lag
-behind the runtime (e.g. 5D tensor support is still rolling out).
+The model is compiled for the GPU accelerator, every signature is run on
+random inputs, and the outputs are compared against a CPU-compiled
+reference. When GPU compilation fails, the runtime's error message names
+the offending op — the patches table in the README maps the common ones
+to the rewrite that clears them.
 """
 
-import collections
 import logging
 
 import numpy as np
 
 log = logging.getLogger("litert_gpu_toolkit")
-
-# Advisory: ops that have historically failed on CompiledModel GPU (ML Drift).
-# Used to suggest patches when GPU compilation fails — not to decide the verdict.
-GPU_SUSPECT_OPS = {
-    'GATHER_ND', 'GATHER', 'SELECT', 'SELECT_V2',
-    'NOT_EQUAL', 'EQUAL', 'GREATER', 'LESS',
-    'TOPK_V2', 'CAST', 'PACK', 'SPLIT',
-}
-
-# Ops that compile to a GPU delegate fallback rather than failing outright.
-GPU_DELEGATED_OPS = {
-    'BATCH_MATMUL',  # YOLO26 C2PSA attention; runs via delegate (verified on Mali-G715)
-}
-
-# Ops that need specific parameter settings
-GPU_CONDITIONAL_OPS = {
-    'RESIZE_BILINEAR': 'align_corners must be False',
-}
-
-# PyTorch modules known to cause GPU issues (for documentation)
-# nn.GroupNorm → ManualGroupNorm (4D reshape approach)
-# Conv2d_WS → bake standardized weights into regular Conv2d
-# F.normalize → manual sqrt+div (div broadcast issue)
-# nn.SiLU/Swish → x * sigmoid(x)
-# nn.GELU → x * sigmoid(1.702 * x)
 
 
 def _static_shape(shape) -> list:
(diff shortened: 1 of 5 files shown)


utilities/litert_gpu_toolkit/checker.py
```diff
@@ -0,0 +1,159 @@
...
+            - 'compatible': bool
+            - 'total_ops': int
+            - 'incompatible_ops': dict of {op_name: count}
+            - 'flex_ops': dict of {op_name: count}
+            - 'warnings': list of str
+            - 'op_distribution': dict of {op_name: count}
+    """
+    import tensorflow as tf
+
+    interp = tf.lite.Interpreter(model_path=tflite_path)
```
smilingday: Can we use CompiledModel API for verification.

smilingday (commented):
Please use new LiteRT CompiledModel API.
We might need to keep udpate or adjust, since some work are in process. For example, we are working on some kind of 5d tensor support, but not sure it's available e2e already.

smilingday:
Another thing is: we can consider to try other devices to verify GPU, beyond Pixel 8, like Qualcomm chip.

Could you check and try this? Request the access:
https://developers.google.com/edge/ai-edge-portal

utilities/litert_gpu_toolkit/checker.py
```diff
@@ -0,0 +1,343 @@
...
+    try:
+        from ai_edge_litert.interpreter import Interpreter
+    except ImportError:
+        try:
+            import tensorflow as tf
+            Interpreter = tf.lite.Interpreter
+        except ImportError:
+            return {}
+
+    interp = Interpreter(model_path=tflite_path)
```
smilingday: Do we still need this?

smilingday (commented):
Seems still have intepreter?

utilities/litert_gpu_toolkit/checker.py
john-rocky: Done — the checker now verifies through the CompiledModel API (47f128b0, then af806d2f): it compiles the model for the GPU accelerator, runs every signature on random inputs, and compares the outputs against a CPU-compiled reference. Thank you for the pointer — this also means the check can't go stale as in-progress runtime work (e.g. the 5D tensor support you mentioned) lands, since the compile result is the source of truth.

utilities/litert_gpu_toolkit/checker.py
john-rocky: No — removed in af806d2f, together with the rest of the interpreter-based op scan. Verification is now purely CompiledModel; when a GPU compile fails, the runtime's error already names the offending op, and the README's patches table maps it to the rewrite.

john-rocky:
Thank you for the review! Status in one line: the checker now verifies purely through the CompiledModel API, with no interpreter usage left (47f128b0, then af806d2f).

Details, if useful:
- It compiles the model for the GPU accelerator (strict GPU first, then GPU+CPU, with the fallback recorded as a warning), runs every signature on random inputs, and compares outputs against a CPU-compiled reference (rtol/atol default 1e-2, since fp16 accumulation makes bit-exactness unrealistic).
- The static op lists are gone, so nothing in the checker will go stale as in-progress runtime work such as 5D tensor support lands — the compile result is the source of truth.
- The API usage follows the patterns already in this repo (`end_to_end/imagenet`, the Qwen3-TTS pipeline). I haven't yet exercised the new checker end-to-end on this machine; I'll run it on the next model conversion and fix anything that differs. Happy to adjust the usage or the tolerances to whatever you prefer.

john-rocky:
On device coverage (re: verifying beyond Pixel 8, e.g. Qualcomm) — agreed, and thank you for the AI Edge Portal link. I'll request access and use it to verify on Qualcomm devices alongside the local Pixel 8 runs, and I'll fold in results from additional hardware as it becomes available.

smilingday:
Thanks you, it's very helpful!

smilingday:
This is already merged?

smilingday:
Can you resove the commens and repush? Then I can approve.
Seems I'm the last pusher or your comments are not addresse, thus cannot approve

john-rocky:
Thank you for the quick follow-up! Both done: the two review comments are now marked resolved (both were addressed in 47f128b0 and af806d2f), and I've repushed the branch — e036f158 merges the latest main in cleanly, no conflicts.

On "is this already merged?" — not yet. What's merged is the gpu-clean-conversion skill (#246, aligned with this checker in #253) and utilities/common (#247). This PR is the remaining piece: the standalone Python toolkit under `utilities/litert_gpu_toolkit` that the merged skill points to as its in-repo implementation.

john-rocky:
Status in one line: not merged yet (main's `utilities/` has only `common` and `tools` — this PR is the one adding `litert_gpu_toolkit`), both review comments are addressed and the branch is freshly pushed, showing MERGEABLE.

Details, if useful:
- "Use CompiledModel API" → addressed in 47f128b0 (checker verifies by compiling for the GPU accelerator, running every signature, and comparing against a CPU reference).
- "Seems still have interpreter?" → addressed in af806d2f (interpreter-based op scan removed entirely; both threads are marked resolved).
- Latest push is e036f158, which merges today's main (including the skills PRs that reference this toolkit) — that may be what looked like "already merged". Ready whenever convenient, and happy to adjust further.

john-rocky:
Found the cause of the approval block and fixed it. My earlier sync merged main into this branch, which pulled your #253/#257/#258/#259 merge commits into the PR's latest push — GitHub then treats you as an author on that push and blocks your approval. I've rebased the branch onto main instead and force-pushed: the tree is identical (no code changes), but the PR now contains only my three commits (head c9841413), so approving directly should work now. If inviting a colleague is still easier, that of course works too.

smilingday (approved):

PR #246, opened 2026-07-29
Add gpu-clean-conversion agent skill

john-rocky:
A first entry for `skills/`, building on #245.

It encodes the conversion loop rather than a single model: convert plain first, gate on the flatbuffer checker, map each symptom to a rewrite in `utilities/litert_gpu_toolkit`, then verify numerically on device. A symptom table does the routing (`GATHER_ND` → which source pattern produced it, rank > 4 → which rewrite, output wrong or NaN → the fp16 reduction family), so an agent can work from what it observes instead of from a guess.

The rule it is built around is that full GPU residency does not imply correct output. Several of the rewrites in #245 exist for cases where the delegate reports N/N nodes and still returns wrong numbers, so the device-side numerical check is a required step in the skill rather than an optional one — along with the tell that distinguishes a silent CPU fallback from a real GPU run (bit-identical to CPU fp32 means it did not run on the GPU).

The output layout follows `models/<family>/<model>/converted/` as landed for Qwen3-TTS, with export, reference dump and verification kept separate so each can be re-run alone. It also treats producing a verified model as the first deliverable, with the demo a later step that needs reusable inference code more than a full UI.

`skills/` is new, so I kept this to one skill rather than guessing at a structure. Happy to reshape the format, or the frontmatter, to whatever convention fits.

### DIFF
--- a/skills/gpu-clean-conversion/SKILL.md
+++ b/skills/gpu-clean-conversion/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: gpu-clean-conversion
+description: Convert a PyTorch or Hugging Face model into a LiteRT model that runs fully on the GPU delegate with verified-correct output, and lay it out as a model recipe. Use when converting a new model, or when a converted model is rejected by the delegate, falls back to CPU, or returns wrong numbers on device.
+---
+
+# GPU-clean conversion
+
+A conversion is done when three things hold, in this order:
+
+1. it converts,
+2. every node runs on the GPU delegate,
+3. **the on-device output matches the source model.**
+
+Step 3 is not implied by step 2. The delegate can report full residency and
+still return wrong numbers — that is the failure mode most of the rewrites
+below exist for. Never call a model done without a numerical check against
+CPU or PyTorch on the actual device.
+
+## Loop
+
+**1. Convert plain first.** No patches. This tells you what the model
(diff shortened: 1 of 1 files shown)


smilingday (approved):
The idea: the skill is not only used by ourselves in this repo, but also by other tools, and for their own models or their own workflow.

john-rocky:
Thank you for the approval, and for spelling out the intent — that the skill should serve other tools and people's own models and workflows, not just this repo. I've updated the skill to match in 8fe632df; nothing is asked of you, and happy to adjust further if the framing misses the mark.

Details, if useful:
- Step 2 now describes the checker as it is after the #245 update: compile for the GPU accelerator via the CompiledModel API, run every signature, compare against a CPU reference — and the symptom table routes on the compile error rather than on an op list.
- Added how to use the toolkit from outside this repo: it is plain Python with no build step, so clone litert-samples and put `utilities/` on `PYTHONPATH`, or vendor the directory.
- The `models/<family>/<model>/` layout is now framed as this repo's convention; the portable rule is the three separately-runnable pieces (export / reference dump / verify) wherever the models live.

john-rocky:
A small correction to my comment above: this merged just before 8fe632df was pushed, so that commit didn't make it into main here. The same change is now #253 — nothing else in it.

PR #260, opened 2026-07-30
utilities/common: reference the androidResources DSL in CompiledModelRunner KDoc

john-rocky:
One-line doc fix in `CompiledModelRunner.kt` (#247): the `fromAssets` KDoc pointed at `aaptOptions { noCompress += "tflite" }`, which is the deprecated name for the block — current AGP spells it `androidResources { noCompress += "tflite" }`, and that is the form already used in this repo (`samples/litert/qualcomm/object_detection/efficientdet/kotlin_npu`). No code change.


smilingday:
I cannot approve somehow. Maybe need to sync?

john-rocky:
Thanks for checking! You were right — the branch had fallen behind main. Rebased onto the latest main and pushed; it should be ready for review now.

smilingday (approved):

PR #261, opened 2026-07-30
Add image_generation sample: Bonsai Image 4B text-to-image apps (Android + iOS)

john-rocky:
Follow-up to #244, adding the on-device apps Lu asked for in the collaboration space (smallest model set, int4 DRQ text encoder), for both platforms.

**What this adds** — `samples/litert/image_generation/`:
- `android/`: Kotlin app on the LiteRT `Interpreter` API (Maven `com.google.ai.edge.litert:litert:2.1.3`), CPU/XNNPACK.
- `ios/`: SwiftUI app on the classic TFLite C API exported by the LiteRT framework, with the XNNPACK delegate attached explicitly — on this path the delegate is a correctness requirement for blockwise int4, not just speed (reference kernels are both ~100x slower and numerically wrong).
- Both carry ports of the full host pipeline: byte-level-BPE Qwen3 tokenizer (token-exact against the Python `transformers` tokenizer on the 26-case golden set in `testdata/`, incl. CJK/emoji/NFD edges), FlowMatch-Euler sampling, and the latent unpatchify. They share a bit-identical seeded-noise stream, so the same (prompt, seed, steps) draws the same image on both platforms (measured 34 dB apart at seed 7; the residual is per-CPU XNNPACK accumulation order).
- The three graphs load sequentially and are freed between stages, keeping peak memory ~2.9 GiB (DiT-sized) instead of the ~4 GiB sum — this is what lets an 8 GB phone finish the run.

**Measured** (smallest set, 3.97 GiB on disk, 512x512, 4 steps):
| Device | DiT step | total |
|---|---|---|
| iPhone 17 Pro | 12–13 s | ~62 s |
| Pixel 8a (8 GB) | 78–88 s | ~7–8 min |

**Verification**: Android `./gradlew testDebugUnitTest` (5/5: tokenizer golden, sigma schedule, position ids, Euler+unpatchify vs recorded fixtures, noise-stream cross-platform goldens) + `assembleRelease`; iOS `xcodebuild` Release. Both apps ran end-to-end on the devices above; model files download from the [model card](https://huggingface.co/litert-community/Bonsai-Image-ternary-4B), which links back to this repo.

I placed it under `samples/litert/image_generation/` to match the existing task-directory layout — happy to move or restructure to whatever location/convention you prefer, and to adjust the license-header attribution if you'd like a different form.


john-rocky:
Thanks for the careful look, Shuangfeng — agreed on both counts, and done in e6e43d92: the directory is now a single README that links the merged python sample, the model card, and the demo apps in my repo, with no app code landed here. The Android numbers stay in my repo's README rather than being featured here, and every vendored header is gone from the PR.

On why the C API headers were copied — a limitation of my setup, not something I'd want upstreamed: the app links the prebuilt CLiteRTLM.framework from the LiteRT-LM releases, which exports the classic TfLiteInterpreter C symbols but ships no headers for them, so I declared the handful I use locally. I took the classic path in the first place because in the v0.13.1 binaries I couldn't find a public thread control on the CompiledModel CPU path (measured single-threaded ~38.8 s per DiT step, vs ~13 s with 6 threads through the classic API + an explicitly attached XNNPACK delegate). If the current runtime works differently I'd love to know.

The bazel-built litert/swift:CLiteRT package looks like exactly the right base — I'll rebuild the app on it following the image_segmentation sample, which drops all of the vendored headers, and bring a polished iOS/macOS demo back here once it's in shape.

smilingday:
I assume this is still old change and need to refresh before review?

john-rocky:
Thanks — just rebased onto the latest main. This is the README-only version (the app code was dropped as we discussed), so it should be ready for review now.

samples/litert/image_generation/README.md
```diff
@@ -0,0 +1,26 @@
...
+The demo apps on both platforms, generating from the same prompt and the same seed — the apps share a bit-identical seeded-noise stream, so seed 7 draws the same red panda on an iPhone and a Pixel:
+
+| iOS (iPhone 17 Pro, ~62 s) | Android (Pixel 8a) |
+|---|---|
+| <img src="img/ios_seed7.png" alt="iOS demo app, seed 7" width="300"> | <img src="img/android_seed7.png" alt="Android demo app, seed 7, with timing log" width="300"> |
+
+## Where things live
+
+- **Python host sample** (tokenize + FlowMatch-Euler loop + unpatchify, ~150 lines): [`models/bonsai/bonsai_image_4b`](../../../models/bonsai/bonsai_image_4b) in this repo.
+- **Demo apps (iOS SwiftUI + Android Kotlin)**: [hf-to-litertlm/bonsai_image_work/device](https://github.com/john-rocky/hf-to-litertlm/tree/main/bonsai_image_work/device) — single-screen apps with on-device Qwen3 tokenization (byte-level BPE in Swift/Kotlin, token-exact against the Python tokenizer), seed control, progress, cancel, and share. A polished iOS/macOS demo app built on the LiteRT Swift package is planned for this repo.
```
smilingday: This needs to be updated? since you already add iOS/macOS demos.

If your android demo app is reasoanble fast, could add here too; otherwise, could link to your personal repo first.

smilingday (commented):
need to update readme.

samples/litert/image_generation/README.md
john-rocky: Thanks for catching this — updated in 19c4924. The README now points the iOS and macOS demos at the in-repo apps from #264 (the `ios/` and `macos/` directories next to it), and per your second option keeps the Android app as a link to my repo for now: on a Pixel 8a a 4-step 512×512 run takes ~7 minutes (78–88 s per DiT step — numbers now in the Measured table), which is more proof-of-run than sample-quality. Happy to bring the Android app into this repo as a follow-up if you'd like it here anyway. Resolving this thread so it doesn't hold up the merge — glad to reopen if you'd prefer otherwise.

smilingday (approved):

PR #265, opened 2026-08-04
Add text_to_speech_streaming sample: KittenTTS nano, dynamic-length streaming TTS (Android)

john-rocky:
This adds the small streaming TTS Android example from last week's thread: KittenTTS nano 0.8 (15M params, 32 MB fp16, 8 voices, Apache-2.0) running fully offline on **dynamic-sequence-length** LiteRT graphs, with sentence-level streaming playback and live **time-to-first-audio / real-time-factor** metrics on screen. Verified on a Pixel 8a: TTFA ≈ 370 ms, RTF ≈ 0.22–0.29 (CPU/XNNPACK, 4 threads). Nothing here needs a decision from you beyond normal review — naming, placement, and the root-README line are all easy to change to whatever fits the repo best.

Details, if useful:

- **Location**: `samples/litert/text_to_speech_streaming/` (sibling of the Matcha `text_to_speech` sample) — Kotlin app + `conversion/` scripts.
- **Dynamic length**: the model is re-authored in TF/Keras so the TFLite converter emits fused dynamic-length LSTM kernels for its five BiLSTMs; the app drives the classic Interpreter API with `resizeInput` per sentence. One graph set runs any text length — no padding buckets, compute scales with the sentence. (This is also why it uses the Interpreter API rather than CompiledModel: the graphs are genuinely dynamic.)
- **Streaming**: sentence-level producer/consumer — sentence N plays while N+1 synthesizes. This granularity is exact (identical waveform to non-streamed output) and matches the upstream pip package. TTFA is one short sentence.
- **Text frontend**: reuses the Matcha sample's espeak-free G2P unchanged (KittenTTS is trained on espeak IPA over the same 178-symbol StyleTTS2 table), with punctuation tokenized the way the upstream pip package does. No GPL components at runtime; fully offline (airplane mode works).
- **Reproducibility**: `conversion/` rebuilds the shipped fp16 graphs **byte-identically** (`build_kitten_tflite.py --fp16`), with goldens, per-stage checks, and an end-to-end verification script. Accuracy sits inside the reference ONNX's own run-to-run noise floor; predicted durations are bit-identical.
- **Models are not committed**: `install_to_device.sh` pushes the five files (three synthesis graphs, the repacked voice table, and the same DeepPhonemizer G2P graph the Matcha sample uses). If it would be convenient, I'm happy to publish the converted `.tflite`s to `litert-community` on Hugging Face (or wherever you prefer) and point the README there.
- The root README "What's New" line is included for convenience — feel free to drop it if you'd rather batch those updates.


john-rocky:
Pushed one fix found in continued device testing (608571e2): the fused dynamic-length LSTM kernels keep their hidden state in variable tensors, and a same-length invoke reused the previous sentence's final state — speaking identical text twice corrupted the predicted durations (96 → 66 frames on a Pixel 8a). The app now calls `resetVariableTensors()` unconditionally before each invoke; verified on device (96/96/96 frames across repeats) and bit-exact repeated waveforms in a Python replica of the app's call pattern. Nothing needed on your side — just keeping the review history transparent.

smilingday:
Could you add a python version of this, so that platforms like Raspberry Pi can demo this easily?

For dynamic-length LSTM support, OKR for now, since we will keep TF->LiteRT conversion path. For LiteRT Torch support, we will work on that: cc: @outtanames @sirakiin 

For using Intepreter API, it's not ideally, and better to use CompiledModel API in some way. We have C++ API for resizeInput, but not in Kotlin API, and we will add that. Meanwhile, @jl45G has some workaround suggestion to achieve this and please suggest.

jl45G:
Thanks Shuangfeng.

HI @john-rocky, could you add a jni binding over litert's reshape https://github.com/google-ai-edge/LiteRT/blob/main/litert/cc/litert_compiled_model.h#L1102 for now 

We will also make this api available in future release.

john-rocky:
@smilingday Thank you for the suggestion, and for the plan on dynamic-length LSTM support — the TF->LiteRT conversion path has been working well for these graphs, so that timeline sounds great to us.

A Python version is now in the PR (`python/` — 9ddf7c57): `say.py` + `kitten_tts.py`, runtime dependencies numpy + ai-edge-litert only, so it runs on a Raspberry Pi 5 (64-bit OS) as-is. It prints the streaming timeline (time-to-first-audio, per-sentence RTF) and writes a 24 kHz WAV.

Details, if useful:

- It uses the **CompiledModel API wherever the graph can load on it**: the G2P (fixed shape) and the vocoder — the vocoder through `resize_input_tensor` per sentence, since the Python API already exposes resize.
- The predictor and prosody graphs still need the Interpreter for now: their fused dynamic-length LSTM kernels keep state in TFLite variable tensors, and the CompiledModel model loader currently rejects those at load time (`model_load.cc`: "Variable tensors not yet supported", the TODO references b/365299994). The moment variable tensors load, the remaining two graphs can switch over with the same call pattern — the code is structured so that swap is local.
- Same asset files as the Android app (the host tables are read from the app's assets directory, so there is one copy in the repo), and the same LSTM-state reset the app needs: the Python Interpreter path calls `reset_all_variables()` before every invoke.
- Verified against the conversion's Mac fp32 references: predicted durations bit-identical on all reference sentences, log-spectrogram correlation 0.984–0.989 (the fp16 bar), repeated synthesis bit-identical. Mac M-series overall RTF ≈ 0.03.

john-rocky:
@jl45G Thank you for the pointer to `ResizeInputTensor` — the binding is in the PR and working on device. Summary: the JNI binding over the C API's resize is added (af22219b), the conv-only vocoder now runs on the CompiledModel API resized to every sentence's length, and the app is re-verified on a Pixel 8a (TTFA 432 ms, RTF 0.291 — unchanged from the Interpreter path). Two graphs could not move yet for reasons unrelated to resize; details below in case they are useful for the official API work. Nothing here blocks us — the current state is a good place for this sample until the official Kotlin API lands.

Details, if useful:

- **What the binding covers** (`cpp/dynamic_shape_jni.cc` + `LiteRtDynamicShape.kt`, resolved from the AAR's libLiteRt.so via dlsym): `LiteRtCompiledModelResizeInputTensor`, `LiteRtGetCompiledModelOutputTensorLayouts(update_allocation=true)`, and a self-contained "run with caller-provided shapes" used for the vocoder (next bullet). One part is necessarily version-pinned: the Kotlin `handle` field holds a `CompiledModelWrapper*` rather than the C handle, so the binding dereferences the wrapper's first member to reach the `LiteRtCompiledModel` — commented in the source and pinned to litert 2.1.5. Happy to adjust any of this if a different shape is more convenient as a reference for the official API.
- **Finding 1 — variable tensors gate the other two graphs.** The predictor and prosody graphs keep fused-LSTM state in TFLite variable tensors, and the CompiledModel model loader rejects those at load time (`model_load.cc`: "Variable tensors not yet supported", b/365299994), so resize alone cannot move them off the Interpreter yet. They switch over the moment that lands.
- **Finding 2 — TFLite-dynamic outputs need caller-sized buffers.** The vocoder's waveform output is a dynamic tensor (its shape only materializes during invoke), so after `ResizeInputTensor` the output-side buffer requirements still reflect the placeholder shape: `LiteRtGetCompiledModelOutputTensorLayouts(update_allocation=true)` returns the placeholder (observable on a Pixel 8a with the 2.1.5 AAR), and buffers created from `createOutputBuffers` are sized from it. `Run` then writes the full-size output into the small registered buffer — observed as a memcpy past the buffer end (SIGSEGV). The same sequence reproduces with the Python API on ai-edge-litert 2.1.6 (`resize_input_tensor` → `create_output_buffers` → `run`). If it would help, I can put together a minimal repro — it may be worth a size re-check at buffer registration so this fails cleanly instead.
- **The workaround in the sample**: the JNI `runDynamic` supplies output buffers of the known final size (`LiteRtCreateManagedTensorBuffer`, host memory) — the waveform length is analytic here (600 samples per frame). With that, the vocoder on CompiledModel matches the Interpreter path bit-exactly in the desktop check, and the app's on-device numbers are unchanged.

smilingday:
Thanks for quick response. It's not ideal to have both compiledmodel and interpreter APIs, but keep it for now :-), and we can improve the codebase further.

smilingday (approved):
Let's polish it further in future.

smilingday:
Seems //samples/litert/semantic_similarity/build_from_source:semantic_similarity
build failure, which might block other submissions

@aravindmurali22 could you help address this?

I will go to merge this PR first.

PR #286, opened 2026-08-19
Replace removed litert_api_with_dynamic_runtime target with fine-grained LiteRT C++ API targets

john-rocky:
**What broke.** Upstream LiteRT removed the `litert/cc:litert_api_with_dynamic_runtime` umbrella target together with the whole `litert/cc/dynamic_runtime` directory ([e1502e07](https://github.com/google-ai-edge/LiteRT/commit/e1502e07)). The WORKSPACE fetches LiteRT `main.tar.gz` unpinned, so Build-and-Test-Linux has been red since the 2026-08-18 scheduled run ([run 32125427177](https://github.com/google-ai-edge/litert-samples/actions/runs/32125427177)): `no such target '@@litert_archive//litert/cc:litert_api_with_dynamic_runtime'`.

**The fix.** Replace all 11 references (semantic_similarity ×2, image_segmentation C++ ×8, image_segmentation iOS ×1) with the fine-grained `litert/cc` API targets each binary's sources include (`litert_compiled_model`, `litert_environment`, `litert_model`, `litert_tensor_buffer`, `litert_options`, …). On current LiteRT main these link the runtime through `litert_runtime_proxy`, which defaults to static linkage, so the samples stay self-contained. The GPU/NPU prebuilt `.so` files in `data` are unchanged.

**Verification**, against a fresh fetch of today's LiteRT main (bazel 7.6.1):

- Every dep label this PR adds exists in the archive and is public in OSS (`bazel query`).
- Full transitive loading of the CI-built targets resolves cleanly (`bazel query 'deps(...)'`) — the failure class above is gone. For `cpp_segmentation_npu` the 2.2 GB QAIRT SDK download kept failing on my machine; its LiteRT deps are identical to the `_mtk`/`_google_tensor` variants, which load cleanly, and its `data` labels are untouched.
- The compile check is this PR's CI run: my macOS machine stops at toolchain configuration before reaching LiteRT code on both the host path (`apple_support` crosstool) and the `android_arm64` path (configure.py needs NDK ≤ 25; I have 28).

@smilingday: this unblocks Build-and-Test-Linux for open PRs.

PR #289, opened 2026-08-20
Add SAM 3 image conversion recipe to models/sam3/sam3_image

john-rocky:
## Summary

New model family under `models/`, following the recipe-first lane of #243/#244/#250/#276/#287: the SAM 3 image-side conversion recipe lands in `models/sam3/sam3_image/converted/` — conversion scripts plus a README carrying the verification numbers. SAM 3 ([facebookresearch/sam3](https://github.com/facebookresearch/sam3), SAM License) takes a text phrase instead of a fixed label set and returns every matching instance with a box, a score and a mask: ViT-L/14 trunk at 1008², a CLIP-L text tower and a text-conditioned DETR head, ~830M parameters, split into three graphs along the data flow. Model producing first: no demo app in this PR.

## What's included

- `sam3_recipe.py` — the self-contained re-authoring recipe. litert-torch silently mis-lowers the >4-D ViT attention (corr 0.607 against torch), so the trunk is rebuilt at ≤4-D: tiled absolute position baked into a buffer, qkv sliced along channels, and the interleaved RoPE de-interleaved by permuting the qkv weight rows, which is exact because q·k is invariant under a shared channel permutation. The DETR decoder is batch-first rank-4 end to end, masked softmax uses the delegate-safe form, ConvTranspose necks become zero-stuff + Conv2d, and SafeLayerNorm carries the |x|≈300 residual stream — full table in the README.
- `build_sam3_image.py` — stock-versus-re-authored parity gate, litert-torch conversion, static op-check by flatbuffer parse, fp16 quantization scoped to matmul-class weights, then CompiledModel CPU parity per graph. Emits the three graphs plus the host-side token-embedding table and CLIP BPE files.
- `verify_sam3_image.py` — real-image end-to-end run through the CompiledModel Python API, with the detection-level comparison against the official PyTorch model when `sam3` is installed.
- `converted/README.md` — environment, pipeline, the re-authoring table, and one finding that generalizes past this model: **the conditioning encoder is what breaks first in fp16, and no aggregate metric shows it.** The CLIP-L text tower's residual reaches |x|≈1.2e3; in fp16 on the GPU the graph compiles, reports fully accelerated, and returns a confident empty result for *some phrases only* — "wheel" keeps its 4 detections while "window" and "paper bag" drop to 0. A single-prompt smoke test passes. The fix is to run that graph on CPU, or on Apple platforms with `enforce_f32`; both are exact and the vision and head graphs stay fp16.

## Verification

- Per graph, fp32 tflite vs the stock PyTorch modules: corr **1.00000** (vision max abs diff 1.9e-4, head 1.8e-3). Static op-check GPU-clean on all three.
- Detection level on real photos, GPU fp16 vision features vs fp32: same kept set, box IoU ≥ 0.993, mask IoU ≥ 0.974 (mean 0.991), |Δscore| ≤ 0.001; a phrase that matches nothing stays empty (presence 0.06). Feature correlation alone misleads here — fp16 features sit at corr 0.994 yet decode to identical detections, so this model is judged at the detection level.
- On device: Pixel 8a (LiteRT 2.2.0, ML Drift) vision `3104/3104` nodes in 1 partition at 9.2 s, head `2791/2791` at 1.4 s, text 0.5 s on XNNPACK — first prompt ≈11 s, re-prompt on the same photo **1.9 s**. iPhone 17 Pro (Metal): ≈5.7 s and **1.3 s**. The split exists for that second column: vision features depend only on the image, so a new phrase re-runs text and head alone.
- **Runtime requirement: LiteRT ≥ 2.2.0.** On 2.1.5 the head graph is mis-executed on the Android GPU — every one of the 200 query logits comes back identical — and the text graph only partially delegates to XNNPACK (1337/1627 nodes, 1.9 s versus 1627/1627 and 0.5 s).
- Checkpoint: `sam3.1_multiplex.pt`, gated on Hugging Face, SAM License; only the `detector.` half is used here. Published deployment files: [mlboydaisuke/SAM3-LiteRT](https://huggingface.co/mlboydaisuke/SAM3-LiteRT).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/models/sam3/sam3_image/converted/.gitignore
+++ b/models/sam3/sam3_image/converted/.gitignore
@@ -1,0 +1,4 @@
+*.tflite
+*.bin
+*.log
+__pycache__/
--- a/models/sam3/sam3_image/converted/README.md
+++ b/models/sam3/sam3_image/converted/README.md
@@ -1,0 +1,121 @@
+# SAM 3 Image Conversion
+
+Model recipes, instructions, scripts, and utilities for converting SAM 3 text-prompted detection and instance segmentation to LiteRT.
+
+These scripts convert the image side of [facebookresearch/sam3](https://github.com/facebookresearch/sam3) (SAM License; ~830M params — ViT-L/14 trunk, CLIP-L text tower, DETR head) into **three `.tflite` graphs that run on the LiteRT CompiledModel GPU**, and numerically verify every step against the PyTorch reference. The published artifacts live at [mlboydaisuke/SAM3-LiteRT](https://huggingface.co/mlboydaisuke/SAM3-LiteRT).
+
+SAM 3 takes a text phrase instead of a fixed label set: type "wheel" or "paper bag" and it returns every matching instance with a box, a score and a mask. The split into three graphs follows the data flow — image features depend only on the image, the prompt embedding depends only on the phrase, and only the head needs both. Caching the vision features per image is what makes a second prompt on the same photo an order of magnitude cheaper than the first.
+
+## Pipeline
+
+```
+image [1,3,1008,1008]  (bilinear to 1008², (x/255-0.5)/0.5, NCHW RGB)
+  ──> [GPU Graph: vision] ──> fpn288 [1,256,288,288] | fpn144 | fpn72   (27.9M floats)
+
+prompt "wheel"
+  ──> host: CLIP BPE (ctx 32, BOS 49406, EOT 49407, zero pad)
+      ──> host: fp16 embedding table lookup ──> emb [1,32,1024]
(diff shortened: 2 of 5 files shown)


smilingday (approved):

john-rocky:
Thanks for the approval. I rebased onto main (now at 54fdec3e). The earlier CI failure came from `samples/litert/semantic_similarity/build_from_source/BUILD` referencing the removed `litert_api_with_dynamic_runtime` target, fixed on main by #286; nothing under `models/sam3/` was involved.

PR #301, opened 2026-09-03
Add SAM 2.1 Hiera-Tiny video-tracking Tensor API implementation (five-signature video stack, host tracking loop, verification suite)

john-rocky:
Thanks, Ping, for the go-ahead in the "the C++ Tensor API" doc thread — "we could use it for some of examples and blogs". This is the SAM 2.1 Hiera-Tiny video-tracking stack authored with the C++ Tensor API, at `models/sam2/sam2_hiera_tiny_video/tensor_api/`, in the same layout as the Qwen3-TTS suite (#241). No weights are included, and nothing here is time-sensitive; happy to reshape any of it to your conventions.

**What this adds**

- `sam2_video/` — the video stack as five signatures in one flatbuffer sharing weights: `encode` (Hiera at the native 1024, raw top-level features plus the two high-res skips), `memcond7` / `memcond2` (memory attention over a fixed bank of 7 or 2 spatial memory slots + 64 object-pointer tokens, unused entries masked additively), `decode` (video mask decoder: all four mask tokens, IoU scores, object pointers, object score), `memorize` (mask downsampler + ConvNeXt fuser memory encoder). Alongside: the weight loader and the host tracking loop (`sam2v_main.cc`), which mirrors the numpy specification in `verify/verify_video_1024.py`.
- `sam2_image/` — the image-path library the video graphs build on (Hiera encoder, in-graph prompt encoder, mask decoder), plus its standalone 512 sample and PyTorch parity script.
- `sam2_video/verify/` — fp32 weight export from `facebook/sam2.1-hiera-tiny` (conv layouts pre-permuted, RoPE bake self-checked in-run), the `transformers` `Sam2VideoModel` streaming reference, per-frame compare, per-signature isolation probes, and a block-by-block encoder mirror.
- No new op class anywhere: SAM 2's pairwise-interleaved RoPE becomes the rotate-half form by permuting the q/k projection rows at weight export (q'·k' == q·k exactly), with the cos/sin tables baked as constants.

**Verification**

- Chained 10-frame clip vs the HF streaming reference: min mask-IoU 1.0000 on CPU fp32 and Metal fp32 (both bank sizes), 0.9950 at Metal fp16.
- M4 Max Metal fp16, per tracked frame end to end (host loop included): 94 ms with the 7-slot bank, 64 ms with 2 slots. Per-graph breakdown and the fp32 / CPU numbers are in the README.
- Validated at LiteRT `a19d8fa` + the PR #8796 runner change.

**Build**

BUILD files target this repository's `@litert_archive` workspace in the same form as #241 (no `litert_api_with_dynamic_runtime`). From the repo root:

```
bazel build --config=macos_arm64 //models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video:sam2v_main
bazel build --config=macos_arm64 //models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image:sam2_main
```

The numbers above come from the same sources built as a `tensor/examples/` overlay in a LiteRT checkout. In this workspace the dependency closure of both binaries resolves under `bazel query`; a `--config=macos_arm64` build stops in analysis at the apple_support crosstool target (the `.bazelrc.user` issue noted on #241, still present on `main`), so the compile in this workspace is not exercised on my side. If anything drifts against current `main`, tell me and I will chase it.

**Relation to #283**

#283 adds the parent directory's README and the `converted/` recipe (the litert-torch export of the same four per-frame graphs). This PR adds only `tensor_api/`; the file sets are disjoint, so the two merge in either order. The README here names `converted/` as its companion.

**Provenance**

Sources are the same as [john-rocky/litert-tensor-vision-examples](https://github.com/john-rocky/litert-tensor-vision-examples) (`sam2_video/`, `sam2_image/`), adapted to this layout; the findings ledger stays there. The memory bank rides host-side as per-frame signature I/O; moving it in-graph as signature state (`odml.cache_update` + the #8796 feedback-loop runner) is the intended next step, and the README records what blocks it on Metal today.

Happy to add whatever an example or a blog post needs — a shorter walkthrough, figures, a trimmed sample — and to adjust placement or naming to whatever fits the repo best. Thanks again for the pointers through this thread; they are what let the video stack drop in cleanly.

### DIFF
--- a/models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image/sam2_graph.cc
+++ b/models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image/sam2_graph.cc
@@ -1,0 +1,853 @@
+// SAM2.1 hiera-tiny image path — graph construction. See sam2_graph.h.
+
+#include "models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image/sam2_graph.h"
+
+#include <cmath>
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "absl/log/absl_check.h"  // from @com_google_absl
+#include "absl/strings/str_cat.h"  // from @com_google_absl
+#include "flatbuffers/flexbuffers.h"  // from @flatbuffers
+#include "tensor/arithmetic.h"
+#include "tensor/buffer.h"
+#include "tensor/datatypes.h"
+#include "tensor/tensor.h"
+
+namespace litert::tensor::examples::sam2 {
+
+namespace {
+
+constexpr float kTwoPi = 6.283185307179586f;
+
+TfTensor ConstScalar(float value) {
+  return TfTensor({.type = Type::kFP32,
+                   .shape = {1},
+                   .buffer = OwningCpuBuffer::Copy<Type::kFP32>({value})});
+}
+
+TfTensor ConstFloats(const std::vector<float>& values,
+                     const std::vector<int>& shape,
+                     const std::string& name = "") {
+  return TfTensor({.name = name,
+                   .type = Type::kFP32,
+                   .shape = shape,
+                   .buffer = OwningCpuBuffer::Copy<Type::kFP32>(values)});
+}
+
+const TfTensor& W(const WeightMap& weights, const std::string& name) {
+  auto it = weights.find(name);
+  ABSL_CHECK(it != weights.end()) << "missing weight: " << name;
+  return it->second;
+}
+
+// Reads a loaded fp32 weight back to host floats (for build-time baking).
+std::vector<float> HostFloats(const TfTensor& tensor) {
+  auto buffer = tensor.GetBuffer();
+  ABSL_CHECK(buffer.ok()) << "weight has no buffer";
+  auto lock = buffer->Lock();
+  const float* data = reinterpret_cast<const float*>(lock.data());
+  return std::vector<float>(data, data + lock.size() / sizeof(float));
+}
+
+// NHWC zero padding on H and W.
+TfTensor PadHW(const TfTensor& x, int top, int bottom, int left, int right) {
+  TfTensor paddings(
+      {.type = Type::kI32,
+       .shape = {4, 2},
+       .buffer = OwningCpuBuffer::Copy<Type::kI32>(
+           {0, 0, top, bottom, left, right, 0, 0})});
+  return Pad(x, paddings);
+}
+
+// Emission toggles (single-threaded graph construction, mirrored from the
+// talker example's pattern).
+bool g_layer_norm_composite = false;
+bool g_sdpa_composite = false;
+bool g_rbmm_attention = false;
+bool g_rbmm_hypernet = false;
+
+// odml.runtime_bmm control inputs for the CURRENT Build*() call, one per
+// distinct bound length S. The ml_drift parser requires the control tensor
+// to be a RUNTIME input (GetNumberOfRuntimeInputsForNode != 3 rejects the
+// node outright), so these are graph inputs, not constants.
+std::vector<std::pair<int, TfTensor>> g_rbmm_params;
+
+TfTensor GetRbmmParam(int s) {
+  for (const auto& [len, tensor] : g_rbmm_params) {
+    if (len == s) return tensor;
+  }
+  TfTensor param({.name = absl::StrCat("rbmm_s", s),
+                  .type = Type::kI32,
+                  .shape = {1, 1, 1, 7}});
+  g_rbmm_params.emplace_back(s, param);
+  return param;
+}
+
+std::vector<uint8_t> RuntimeBmmAttributes(bool is_src) {
+  flexbuffers::Builder fbb;
+  fbb.Map([&]() {
+    fbb.Bool("is_global", true);
+    fbb.Bool("is_src", is_src);
+    fbb.Bool("rhs_cache_update", false);
+  });
+  fbb.Finish();
+  return fbb.GetBuffer();
+}
+
+// The runtime_bmm contract (parser: transpose_right=true): LHS [B,H,M,K] x
+// RHS [B,H,N,K] -> [B,H,M,N]; the batch dim merges into H on the delegate
+// when B > 1 (window attention). Decompositions are plain BatchMatMuls that
+// ignore the control tensor, so CPU execution equals the raw path.
+TfTensor RuntimeBmm(const TfTensor& lhs, const TfTensor& rhs,
+                    const TfTensor& param, bool is_src) {
+  StableHLOCompositeOptions opts{
+      .name = "odml.runtime_bmm",
+      .composite_attributes = RuntimeBmmAttributes(is_src)};
+  return StableHLOComposite(
+      opts,
+      [](TfTensor l, TfTensor r, TfTensor /*p*/) {
+        return BatchMatMul(l, r, /*adj_x=*/false, /*adj_y=*/true);
+      },
+      lhs, rhs, param);
+}
+
+std::vector<uint8_t> EpsilonAttributes(float eps) {
+  flexbuffers::Builder fbb;
+  fbb.Map([&]() { fbb.Float("epsilon", eps); });
+  fbb.Finish();
+  return fbb.GetBuffer();
+}
+
+// Last-axis LayerNorm with weight+bias. In NHWC this also covers the
+// reference's LayerNorm2d (channelwise over NCHW == last axis here).
+TfTensor LayerNormRaw(const TfTensor& x, const TfTensor& weight,
+                      const TfTensor& bias, float eps) {
+  int last = static_cast<int>(x.GetShape().size()) - 1;
+  TfTensor mean = Mean(x, {last}, /*keep_dims=*/true);
+  TfTensor centered = Sub(x, mean);
+  TfTensor var = Mean(Mul(centered, centered), {last}, /*keep_dims=*/true);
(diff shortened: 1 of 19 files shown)


models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image/sam2_graph.cc
```diff
@@ -0,0 +1,853 @@
...
+      {.type = Type::kI32,
+       .shape = {4, 2},
+       .buffer = OwningCpuBuffer::Copy<Type::kI32>(
+           {0, 0, top, bottom, left, right, 0, 0})});
+  return Pad(x, paddings);
+}
+
+// Emission toggles (single-threaded graph construction, mirrored from the
+// talker example's pattern).
+bool g_layer_norm_composite = false;
```
pyu10055: instead of using global mutable variable, create a struct to hold the model config.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_weights.cc
```diff
@@ -0,0 +1,191 @@
...
+
+namespace litert::tensor::examples::sam2_video {
+
+namespace {
+
+// The gemma3 loader's Gemma-norm heuristic: +1.0 on any tensor whose name
+// contains "layernorm" or ends with "norm.weight". Three video keys match
+// (memory_attention.norm.weight, memory_encoder.fuser.{0,1}.norm.weight);
+// their loaded values are compensated back below.
+bool LoaderAddsGemmaOffset(const std::string& name) {
```
pyu10055: Add an option flag to SafetensorLoader (e.g., apply_gemma_norm_offset = false) rather than relying on substring matching and mutating buffers post-load.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_main.cc
```diff
@@ -0,0 +1,668 @@
...
+  consts.not_a_point =
+      HostFloatsOf(weights, "sam_prompt_encoder.not_a_point_embed.weight");
+  consts.track_sparse = HostFloatsOf(weights, "tables.track_sparse");
+  consts.mtpe = HostFloatsOf(weights, "maskmem_tpos_enc");  // (7,1,1,64)
+  consts.no_obj_ptr = HostFloatsOf(weights, "no_obj_ptr");
+  consts.tpos_w = HostFloatsOf(weights, "obj_ptr_tpos_proj.weight");
+  consts.tpos_b = HostFloatsOf(weights, "obj_ptr_tpos_proj.bias");
+
+  // ---- Build the five signatures (shared weights, one flatbuffer). ----
+  std::vector<float> real_no_mem = HostFloatsOf(weights, "no_mem_embed");
```
pyu10055: Add a boolean option in Sam2Config (e.g., bool fold_no_mem_embed = true;) so BuildEncoder can conditionally omit the no_mem_embed addition without mutating WeightMap.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_graph.cc
```diff
@@ -0,0 +1,540 @@
...
+
+  // Fuse with the projected pixel features.
+  TfTensor f = Conv2D(inputs.pix_raw, W(weights, me + ".pix_feat_proj.weight"),
+                      W(weights, me + ".pix_feat_proj.bias"), 1, 1,
+                      kPaddingValid);
+  f = Add(f, m);
+
+  // 2 ConvNeXt fuser blocks: depthwise 7x7 (pad3 + VALID), LN, pw1, GELU,
+  // pw2, gamma scale, residual. Depthwise filter repacked to [1,7,7,C].
+  for (int i = 0; i < 2; ++i) {
```
pyu10055, line 498: Since export_weights_1024.py
 already permutes all Conv2D and ConvTranspose2D weights to TFLite layout (OHWI), permuting depthwise conv weights to [1, kh, kw, C] in Python during export would simplify C++ graph construction.

john-rocky:
Thanks for the review, Ping — all four points are addressed, one commit each: [eb8cdf5](https://github.com/google-ai-edge/litert-samples/pull/301/commits/eb8cdf5491eecd7db26b206576578d8f0191f6d0) (build-scoped context instead of globals), [f0ac2b3](https://github.com/google-ai-edge/litert-samples/pull/301/commits/f0ac2b37a5daa09434988b21e5da2c51a212f690) (weights read as stored, no norm-offset compensation), [3cb84ad](https://github.com/google-ai-edge/litert-samples/pull/301/commits/3cb84ad0da759182ea4761a593949537549e212b) (`fold_no_mem_embed`), [07d7e49](https://github.com/google-ai-edge/litert-samples/pull/301/commits/07d7e49862b90f5723ca169e52cb7d8964d5d2df) (depthwise filters permuted at export). Details are in the threads. The one place I deviated from the suggestion is the loader flag: `SafetensorLoader` is LiteRT code, so this sample now reads tensors through the loader's metadata instead; happy to add the flag upstream if you prefer that route.

Re-verified at the new head with a fresh weight export and a fresh HF streaming reference (same LiteRT pin as the description, M4 Max):
- chained 10-frame parity vs `Sam2VideoModel`: min mask-IoU 1.0000 on CPU fp32 and Metal fp32 for both bank sizes, per-frame foreground counts equal to the reference on every frame; Metal fp16 0.9952 (the description says 0.9950 — the pre-review binary also measures 0.9952 on this machine today, so the difference is not from these changes);
- per-signature isolation probes on HF-captured inputs: corr 1.000000 for encode / decode / memorize / memcond;
- the 512 image path still matches its PyTorch reference to the printed digits (iou 0.1189 / 0.2172 / 0.9859, object score 23.966), and the raw and rbmm builds agree bit for bit.

Nothing else changed. If anything else would make the import easier, tell me and I'll take care of it — thanks again for the careful read.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_image/sam2_graph.cc
john-rocky: Thanks — agreed, those globals were carried over from the single-binary version. Fixed in [eb8cdf5](https://github.com/google-ai-edge/litert-samples/pull/301/commits/eb8cdf5491eecd7db26b206576578d8f0191f6d0): the four toggles and the runtime_bmm control-input registry now live in one build-scoped `BuildContext` (a reference to the `Sam2Config` plus the control inputs created for that graph), threaded through the LayerNorm / attention helpers. The control inputs come back on `EncoderOutputs` / `DecoderOutputs` as `rbmm_params`, replacing `TakeRbmmParams()`. Graph output is unchanged: at 512 on CPU, the `--attention=rbmm --hypernet=rbmm` build still matches the raw path bit for bit (masks, iou, object score).

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_weights.cc
john-rocky: Thanks for flagging this — the substring match was fragile, and mutating the buffers after load compounded it. `SafetensorLoader` lives in LiteRT (`tensor/examples/gemma3`), so I could not add the flag from this repo; [f0ac2b3](https://github.com/google-ai-edge/litert-samples/pull/301/commits/f0ac2b37a5daa09434988b21e5da2c51a212f690) removes the need for it instead. Both loaders now go through one `LoadWeightSpecs` that opens the file with `SafetensorLoader` and reads each tensor's bytes from its metadata (F32 as stored, F16/BF16 widened), shape-checked against the spec, so no name-based rewriting is left in the sample. If you would like the flag on the loader itself (e.g. `apply_gemma_norm_offset`, default true so Gemma is unaffected), I'm glad to send that as a LiteRT PR and switch this sample back to `LoadTensor` once it lands.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_main.cc
john-rocky: Thanks — done in [3cb84ad](https://github.com/google-ai-edge/litert-samples/pull/301/commits/3cb84ad0da759182ea4761a593949537549e212b): `Sam2Config::fold_no_mem_embed` (default true, the image path's contract). The video host clears it and `BuildEncoder` skips the addition, so the encoder emits the FPN feature map directly and the `WeightMap` is never touched. The video encoder also loses the zero Add it used to carry.

models/sam2/sam2_hiera_tiny_video/tensor_api/sam2_video/sam2v_graph.cc
john-rocky: Thanks — that repack was the one filter the graph still rearranged at build time. In [07d7e49](https://github.com/google-ai-edge/litert-samples/pull/301/commits/07d7e49862b90f5723ca169e52cb7d8964d5d2df), `export_weights_1024.py` writes the two fuser depthwise filters as `[1, kh, kw, C]` (`permute(1, 2, 3, 0)` on torch's `[C, 1, kh, kw]`), the weight spec expects that shape, and `BuildMemorize` passes the stored tensor to `DepthwiseConv2D` directly. The export's cross-check against the 512 file compares them after transposing back: 456 keys, 0 mismatches. The parity numbers in my summary comment are from the re-exported file.

pyu10055 (approved):
thanks for contributing!

PR #304, opened 2026-09-09
Add the model conversion cookbook (models/README.md): step-by-step instructions from a Hugging Face checkpoint to a verified LiteRT bundle

john-rocky:
Thanks for the suggestion in the MiniCPM5-2B thread to add instructions / a cookbook here. Until now the procedure lived in two places that a person converting their own model would not read end to end: the agent skills under `skills/` (written for coding agents, with the facts spread over five reference files) and the MiniCPM5-2B recipe README (one model's story). This PR adds the page in between: `models/README.md`, a step-by-step walk-through from a Hugging Face checkpoint to a verified `.litertlm` bundle, with the recipes under `models/` as its worked examples.

The page runs in the order a conversion does, from classifying the architecture in `config.json` to publishing behind a mechanical gate, and ends with a symptom-to-fix table, the MiniCPM5-2B worked example and a short section on the `.tflite` lane. Every fact on it is already in the skills or the recipe README; the page orders them for a human reader and links to the reference behind each rule. No new measurements: every number on the page is taken from the recipe README or the skill references.

Two small index changes ride along: `litert-conversion-workflow` was missing from `skills/README.md` and from the top-level README's skills list, and the top-level README's `models/` section now points at the cookbook.

Checked before opening: every relative link and anchor on the page resolves on this branch, and the int4 recipe JSON is the recipe `build_minicpm5_2b.py` registers (blockwise-32 + OCTAV on linears, int8 embedding).

If a different location or shape fits the repo better (a `docs/` page, one page per lane), I'm happy to move it. Thanks again for the pointer.

CI note: Markdown only, three README files.


smilingday:
Can you rename README.md as conversion.md or similar?

For this model repo, I expect we add both recipes like how to use the models, examples, as well as detailed conversions.

For README, we should make it simpler for human readable content, that are useful for all models.
E.g. how to use those models, where to find examples, how to convert, and a model list with links for detials. etc.

john-rocky:
Thanks for the review, and for spelling out what the models/ README should be. Done: the cookbook is now models/conversion.md, unchanged, and models/README.md is a 32-line page with the four items you listed: how to run the models (the LiteRT-LM engine and CLI, the CompiledModel API samples, the Tensor API programs), where the examples are, one line to conversion.md, and a model list with each recipe's source model and converted weights. Everything on it is already stated elsewhere in the repo; nothing new is claimed. Per-model usage recipes I'll send as a separate PR, so this one stays the rename. Anything you'd like changed on the page, I'll do right away. Thanks again for the steer.

smilingday (approved):

PR #309, opened 2026-09-13
Add the MiniCPM5-2B usage recipe page (models/minicpm/minicpm5_2b/README.md): run, pick a file, thinking, Python, Android and iOS, serve, tested on

john-rocky:
Thanks again for the steer on #304; this is the per-model usage page I said I would send separately. `models/minicpm/minicpm5_2b/README.md` is now the page a user reads to run the published MiniCPM5-2B bundles, laid out like the per-model recipe pages you pointed at: commands first, then which file to pick, the thinking switch, the Python API, Android and iOS, `litert-lm serve`, a "Tested on" table, and one line to the conversion.

Every command on it is a standard tool of [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM), the LLM runtime of [LiteRT](https://github.com/google-ai-edge/litert), used unchanged. The only model-specific parts are the two file names, the thinking switch and its budget, and the measured numbers, so a new model page copies the sections and changes only those; one paragraph in `models/README.md` fixes that section order, and `models/minicpm/README.md` gets a one-line description. The conversion detail stays in `converted/` and `conversion.md`, and the page's old copy of the failure-mode table is gone, so that table lives in one place.

Every command in the page's code blocks was run today on litert-lm 0.17.0 on a Mac M4 Max. The published files were also run on a Galaxy S26, with the on-device sha256 checked against the manifest; the iPhone 17 Pro row repeats the recipe's earlier result and says that speed was not measured there.

One thing is left off the page on purpose: tool calling. The runtime's template renderer rejects the embedded template's `tojson(ensure_ascii=False)`, and with the canonical MiniCPM 5 template from LiteRT-LM the model's XML function call comes back as plain text; I'll put the two-command reproduction in a LiteRT-LM issue.

Whatever shape or placement fits the repo better, I'll change it. Thanks again.

cc @smilingday


smilingday (approved):
This looks great and much better and useful! Thanks for doing this

john-rocky:
Thank you for the review, Shuangfeng. I've marked the PR ready for review. The one red check is the C++ samples CI, which fails on main at e3a37691 as well (the last three scheduled runs), and this PR only adds the README under models/. Thanks again.

PR #311, opened 2026-09-14
ios/image_segmentation: configure LiteRT CPU options

aravindmurali22:
ios/image_segmentation: configure LiteRT CPU options
- Include litert_cpu_options.h from CLiteRT framework
- Configure delegate kernel mode and thread count when compiling for CPU


junjiang-lab (commented):
Can we use [CpuOptions.swift](https://github.com/google-ai-edge/LiteRT/blob/main/litert/swift/Sources/CpuOptions.swift) instead of "litert_cpu_options.h"?

aravindmurali22:
> Can we use [CpuOptions.swift](https://github.com/google-ai-edge/LiteRT/blob/main/litert/swift/Sources/CpuOptions.swift) instead of "litert_cpu_options.h"?

Hi @junjiang-lab , I have addressed the above comment. Thank you.

junjiang-lab (approved):

PR #312, opened 2026-09-15
Add benchmark/: run LiteRT benchmark_model on Developer Device Platform lab devices

john-rocky:
I run LiteRT benchmarks on Developer Device Platform (DDP) lab phones through the Device Run REST API. This PR adds that as a reference under a new top-level `benchmark/`: one HTTP request per model and device runs the public `benchmark_model` binary, and the results come back as files in a bucket. No app and no adb.

Five files.
- `README.md`: prerequisites, the three steps (upload, pick a request body, submit and fetch), what comes back, how to decode `results.pb`, verification.
- `session-cpu.json`, `session-gpu.json`, `session-cpu-gpu-wrapper.json`: request bodies for CPU, GPU (`--use_gpu=true`), and CPU+GPU in one session through a wrapper script that keeps each run's stdout.
- `multi_benchmark_harness.sh`: that wrapper.

The CPU and wrapper request bodies and the wrapper script are the DDP team's examples. I changed two things: `--export_model_runtime_info=true` is removed from the CPU body (this binary reports it as unconsumed, and `--model_runtime_info_output_file` writes the file on its own), and the script carries the repo's license header. `session-gpu.json` is the CPU body with `--use_gpu=true` added and `-gpu` on its display names. If you would rather credit the script differently, tell me and I will change it.

What I verified. `session-cpu.json` and `session-gpu.json` ran on `pa3q-35` (Galaxy S25 Ultra) with `mobilenet_v2.tflite`: `session-e1d64576` (CPU, XNNPACK, 70/70 nodes) and `session-233c4731` (GPU, OpenCL, 70/70 nodes), both PASSED, each operation done about 60 s after the POST. The pulled `results.pb` decodes and matches the `BENCHMARK RESULTS` block in `logcat.txt`. The wrapper variant ran as shipped on `caiman-35` (Pixel 9 Pro): `session-bd10daf6`, PASSED, with the `output` directory pulled (both runs' logs, `results_<run>.pb`, `runtime_info_<run>.pb`) and no unconsumed-flag warning in either log. The binary's stdout is not stored on this path; the wrapper variant is how to keep it.

Two open points. The wrapper body pushes `binaries/latest/`, a different object from the pinned `2.2.0` the other two use; I can pin all three to one version if you prefer. The root README's Repository Structure list does not mention `benchmark/` yet; I can add two lines in a follow-up once the location is settled.

Draft on purpose: re-point it at whatever shape the DDP examples take, or close it if another shape fits better.

cc @smilingday @ztenghui


smilingday:
this is not needed any more?

john-rocky:
Thanks for checking, Shuangfeng. It is not needed any more: `litert benchmark --ddp` now builds and submits the Device Run request on its own. I'm closing this. The branch stays on my fork in case any of it is useful later. Thanks again.

PR #318, opened 2026-09-17
Add benchmark/leaderboard: a LiteRT performance board across Android, macOS and iOS, with drivers and a skill

john-rocky:
Thanks for the direction in the thread, Shuangfeng: make some PR and iterate from there, other platforms in the schema, static data from local scripts first. This is that PR, as a Draft so you can pass over it before it goes to review: a static board page over committed JSON, the drivers that produce every row, an iOS app runner and a skill, with eight measured rows of MobileNet-v2 across Android, macOS and iOS.

**Why.** While working on conversions I measure the same models again and again, on the lab phones and on my own machines, and the numbers ended up in chat messages. A board fed only by the drivers that produced the numbers makes each row reproducible: a row names the device, the accelerator, the `benchmark_model` build, the session and where each number was read from.

**What is in it.**
- `benchmark/leaderboard/`: the page (`index.html`, the look of the repo's index page) over `data/board.json`, with filters for platform, device, accelerator and task; a click on a row shows every field. `data/measurements.jsonl` keeps every job; `board.json` is the selection the page reads.
- `benchmark/driver/`: `run_matrix.py` submits `litert benchmark --ddp` sessions from `matrix.yaml`; `run_local.py` runs the macOS `benchmark_model` of the same release on the Mac it is started on; `collect.py` decodes `results.pb` with protoc and reads the delegate line from the job log; `build_board.py` keeps the newest row per model, platform, device and accelerator.
- `benchmark/ios/`: an app that builds `benchmark_model` from [LiteRT](https://github.com/google-ai-edge/litert)'s source at the pinned tag and runs it on an iPhone; `run_ios.sh` writes the same session layout, so its rows go through the same `collect.py`.
- `skills/benchmark-on-ddp/SKILL.md` (and its line in `skills/README.md`): the loop for an agent, DDP and local.

A row is (model repo, file) × platform × device × accelerator × `benchmark_model` release; a new release is a new row, and earlier rows stay in `measurements.jsonl`. The board groups rows by platform and, by default, orders them by median latency inside the group; it ranks nothing across platforms.

**Measured.** `mobilenet_v2.tflite` from litert-community/MobileNet-v2 at 2.2.0, CPU and GPU each: Pixel 9 Pro and Galaxy S25 Ultra through DDP on 2026-09-16; a Mac Studio (M4 Max) through `run_local.py` and an iPhone 17 Pro through the app on 2026-09-17. Every row carries its session id and the build it ran; the environment is in the README's "Tested on", and the page's "How the rows are measured" says what the binary did.

**Limits.** The DDP rows depend on the `--ddp` target in review in [LiteRT-CLI](https://github.com/google-ai-edge/LiteRT-CLI); until it lands, the skill installs the CLI from the review branch. Nodes delegated (N/M) is read from the Android device log; the macOS binary's stdout and the app's log have no such line, so those rows show the delegate name only. [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM) bundles are not on the board yet. A job with no results is reported by the driver and never becomes a row.

Whatever should change here, layout, names or which fields are public, I will do it in this PR. Thanks for looking.

cc @smilingday


smilingday:
Overal good draft, and let's get this in and iterate.
For DDP change, already merged and you can use nightly tomorrow (litert-cli-nightly, see whether your ddp change works well).

I cannot approve, since last change is by me.

smilingday (approved):

john-rocky:
Thanks for re-landing the --ddp change and for the nightly pointer. It works from the nightly: with litert-cli-nightly 0.3.0.dev20260918 in a fresh venv, `litert benchmark models/mobilenet/mobilenet_v2.tflite --ddp --device caiman-35 --gcp-project litert-edge-portal` ran one session on caiman-35 (session-ce215032, PASSED, outputs pulled). Thanks again.

PR #320, opened 2026-09-18
benchmark/leaderboard: LiteRT-LM rows from the prebuilt benchmark binary

john-rocky:
Thanks for the LiteRT-LM benchmark binary, Shuangfeng. This Draft adds a second table to the leaderboard: prefill and decode tokens/s, time to first token and init of a `.litertlm` bundle under [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM)'s `litert_lm_advanced_main --benchmark`, one run per backend. It follows #319 (the [LiteRT](https://github.com/google-ai-edge/litert) `benchmark_model` rows), which is in.

- Run: `benchmark/driver/litert_lm_harness.sh` runs the binary from the bucket's `litert_lm/` directory on a DDP phone once per backend with the flags from the review (5 iterations, caches on; the harness removes the cache files beside the bundle first) plus `--metric_proto_file_path`, and keeps each run's metrics proto, log and exit code. The session pushes the binary, `libLiteRtOpenClAccelerator.so` and `libLiteRtTopKOpenClSampler.so` from that directory, `libGemmaModelConstraintProvider.so` from LiteRT-LM's `prebuilt/android_arm64` (the binary links it at start and the directory does not carry it; asked in the review thread), and the bundle. On caiman-35 the log reads "Dynamically loaded GPU accelerator(libLiteRtOpenClAccelerator.so) registered" and the GPU run went through OpenCL.
- Collect: `collect_lm.py` decodes `metrics_<run>.pb` against `litert_lm_metrics.proto` at v0.17.0 (unchanged on main) into `data/measurements-lm.jsonl`; `build_board.py` adds `lm_rows` to `board.json`; the page lists them below the `benchmark_model` table with every field on click. A row shows the median of tokens/s and first-token time over iterations 2 to 5 (the first is a warm-up, `runtime_lm.warmup_iterations`) and the run's cold-start init, and keeps each iteration's numbers. `matrix.yaml` names the binary by sha256 (`adac974b…`, the `latest` object of 2026-09-18) and each library with its source and sha256, the flags and the token counts under `runtime_lm`, and the bundle list under `lm_models`; the row id carries the token counts, so a run at other counts lands as its own row (the short-context condition for now).
- First rows and their limits: litert-community/Qwen3-0.6B (`qwen3_0_6b_mixed_int4.litertlm`) on caiman-35 (Pixel 9 Pro), one session on 2026-09-19, 1024/256 tokens, 5 iterations: GPU (OpenCL) 591 tokens/s prefill and 21.9 decode, CPU 139 and 3.3, the CPU run throttling from its third iteration (prefill 186 to 203 tokens/s in the first two, 92 to 93 in the last two; thermal status 0 before that run and 1 after it) while the GPU run held 584 to 597; GPU init 7.4 s with the caches on (19 s yesterday with them off) (60 s of cooldown between runs). Memory is not collected: the runs did not pass `--report_peak_memory_footprint`. The rows are provisional until every input comes from the directory.

Whatever should change in the layout, the fields or the flags, I will do it here. Thanks for looking.

cc @smilingday


benchmark/driver/matrix.yaml
```diff
@@ -48,6 +48,26 @@ platforms:
...
+runtime_lm:
+  # LiteRT-LM's benchmark binary the LLM rows ran (`latest` moves, so the rows
+  # keep the object's date in the version and its sha256 in the binary field),
+  # the git ref its metrics proto is read at, the flags every run passes and the
+  # token counts of the default condition.
+  version: "latest@2026-09-18"
+  binary: latest/android_arm64/litert_lm_advanced_main
+  sha256: e10edfa46596705db6966506e25548731074f889edac161858b730279778c68c
+  proto_ref: v0.17.0   # the tag whose runtime/proto/litert_lm_metrics.proto decodes the metrics (unchanged on main)
+  flags: [--disable_cache=true, --num_iterations=3]
```
smilingday: suggest to remove 
--disable_cache=true
and run num_iterations=5

the first time for warm up, and then compute the numbers of rest.

benchmark/driver/matrix.yaml
```diff
@@ -48,6 +48,26 @@ platforms:
...
+  conditions: {prefill_tokens: 1024, decode_tokens: 256, max_num_tokens: 1280}
```
smilingday, line 73: We can first get short context benchmarking in using those settings.

Later, we can add long context benchmark, as offline discussions

benchmark/driver/matrix.yaml
```diff
@@ -48,6 +48,26 @@ platforms:
...
+
+lm_models:
+  # repo: Hugging Face repo. file: the .litertlm bundle. task: the repo's
+  # pipeline tag. backends: cpu, gpu (the binary's --backend values).
+  - repo: litert-community/Qwen3-0.6B
+    file: qwen3_0_6b_mixed_int4.litertlm
```
smilingday, line 79: Later, you can reupload the models with newly converted models?

Good to ahead for now.

benchmark/driver/matrix.yaml
```diff
@@ -48,6 +48,26 @@ platforms:
     # which builds benchmark_model from LiteRT's source and runs it on a phone.
     # The session names the phone, its OS and the source tag (see ../ios/README.md).
 
+runtime_lm:
+  # LiteRT-LM's benchmark binary the LLM rows ran (`latest` moves, so the rows
+  # keep the object's date in the version and its sha256 in the binary field),
+  # the git ref its metrics proto is read at, the flags every run passes and the
+  # token counts of the default condition.
+  version: "latest@2026-09-18"
+  binary: latest/android_arm64/litert_lm_advanced_main
```
smilingday: Please use new dirctory:

gs://litert/binaries/latest/android_arm64/litert_lm/embedding_litert_lm_main

There are other files under the same buckets if you want in future:

libLiteRtOpenClAccelerator.so
libLiteRtTopKOpenClSampler.so
embedding_litert_lm_main
litert_lm_main
libLiteRtDispatch_Qualcomm.so
libLiteRtDispatch_MediaTek.so
libLiteRtDispatch_GoogleTensor.so

smilingday (commented):
Left some comments.

In high level, could you build
litert benchmark --ddp <..../litertlm file>
capabilties, so that here, we can reuse that benchmarking capabilties.

We can make that changes too in following up PRs.

benchmark/driver/matrix.yaml
john-rocky: Applied, thanks: `--disable_cache=true` is out and a run is 5 iterations; `collect_lm.py` leaves the first iteration out and takes the medians over the other four (each iteration stays in the row). `litert_lm_harness.sh` removes the cache files beside the bundle before a run (the caches stay on during it), so Init is a cold start each time.

benchmark/driver/matrix.yaml
john-rocky: Understood, thanks: short context only for now, at these settings, noted as such in matrix.yaml; a long-context condition can come later as rows of its own, since the row id carries the token counts.

benchmark/driver/matrix.yaml
john-rocky: Yes, thanks: once the newly converted bundles are ready I reupload them and rerun them here; a bundle with the same file name replaces the row, and a new file name becomes a new row.

benchmark/driver/matrix.yaml
john-rocky: Thanks for the new directory and the list of what it holds; switched. The rows use its `litert_lm_advanced_main` (sha256 `adac974b…`; `embedding_litert_lm_main` has no `--benchmark`, so I took the note as the directory; say if `litert_lm_main` is the one you meant) with its `libLiteRtOpenClAccelerator.so` and `libLiteRtTopKOpenClSampler.so` beside it, which the GPU run loads (without them it exits with no accelerator). One thing I hit: the binary links `libGemmaModelConstraintProvider.so` at start and the directory does not carry it, so on a Galaxy S26 here it stops before the benchmark: `CANNOT LINK EXECUTABLE "./litert_lm_advanced_main": library "libGemmaModelConstraintProvider.so" not found: needed by main executable`. For these rows the session pushed the copy from LiteRT-LM's `prebuilt/android_arm64` (git LFS object `f2aa38ce…`, last changed in 4453b286) beside it and it ran on caiman-35 (session-aba54d60, 2026-09-19); matrix.yaml names each library with its source and sha256, and the rows are provisional until every input comes from the directory. Could the directory carry that library too? Thanks again.

john-rocky:
Thanks for the review and the pointer to the `litert_lm/` directory, Shuangfeng. The four notes are in: caches on with 5 iterations and the first left out of the medians; short context only for now; the reupload once the newly converted bundles are ready; and the directory, where the one snag is that its `litert_lm_advanced_main` needs `libGemmaModelConstraintProvider.so` beside it (details and one question in that thread). The rows are re-measured with these settings on caiman-35 (Pixel 9 Pro), Qwen3-0.6B, 1024/256 tokens, median of iterations 2 to 5: CPU 139 tokens/s prefill and 3.3 decode, GPU (OpenCL) 591 and 21.9. The CPU run throttled from its third iteration (thermal status 0 to 1), which pulls its medians below yesterday's three-iteration numbers; every iteration is in the row. The CLI side is a Draft PR in LiteRT-CLI, https://github.com/google-ai-edge/LiteRT-CLI/pull/88: `litert benchmark <bundle>.litertlm --ddp` runs the same binary from that directory, pulls the metrics proto and prints the medians after the warm-up; end to end once the directory carries that library, a Draft until then. Thanks again.

smilingday:
libGemmaModelConstraintProvider.so is available now.

Btw, if we get litert benchmark works for LM, this harness script can be skipped?
We can follow up later, once that are released.

smilingday (approved):

PR #321, opened 2026-09-18
Add the Bonsai Image 4B usage recipe page (models/bonsai/bonsai_image_4b/README.md): run, pick a file, steps and seed, Python, Android and iOS, tested on

john-rocky:
Thanks again for the review of #309; this is the second model page in that shape. `models/bonsai/bonsai_image_4b/README.md` is now the page a user reads to run the published Bonsai Image 4B graphs, in the same section order: commands first, then which file to pick, the model's own switches (steps, seed and threads), a Python block, Android and iOS, a "Tested on" table, and one line to the conversion.

Bonsai Image 4B is the first `.tflite` page in the set, so two things differ from the MiniCPM5-2B page and are worth a look. There is no Serve section, since the three graphs run through the [LiteRT](https://github.com/google-ai-edge/litert) runtime and not the [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM) engine; `models/README.md` now says Serve is for `.litertlm` bundles. And the Python block is one CompiledModel run on the text encoder, the first stage of the host loop that ships with the recipe, so a reader who ports the loop to Kotlin or Swift sees the three rules that matter: inputs by argument position, one graph resident at a time, XNNPACK on.

Three fixes rode along: with transformers 5.17 the pip line needs `jinja2`, or `generate.py` stops at the chat template, so the line on the page and `python/requirements.txt` gained it; after the review, `python/generate.py` moved from the Interpreter to the CompiledModel API with the same calls as the block, the seed-0 image byte-identical to before; and the source-checkpoint link on this page and in the recipe table now points at `prism-ml/bonsai-image-ternary-4B-unpacked`, the repo the conversion scripts load, since the old name no longer resolves on the Hub. The conversion detail stays in `converted/`.

Every command in the page's code blocks was run today on ai-edge-litert 2.2.0 on a Mac M4 Max, and the macOS sample app was run on the same machine on Metal; those two rows are the "Tested on" table. The iOS app's README keeps its own iPhone 17 Pro measurement, and Android is marked as not measured.

Whatever shape or placement fits the repo better, I'll change it. Thanks again.

cc @smilingday


smilingday (dismissed):
Thanks!

models/bonsai/bonsai_image_4b/README.md
```diff
@@ -1,41 +1,84 @@
...
+python python/generate.py --model-dir bonsai --prompt "a red fox sitting in fresh snow at sunrise" --steps 8 --seed 42 --out fox_8.png
+python python/generate.py --model-dir bonsai --prompt "a red fox sitting in fresh snow at sunrise" --threads 8 --out fox_t8.png
+```
+
+The model is distilled for four steps; `--steps` takes more, at one DiT pass per step. The same prompt, seed and step count give the same image on the same machine, whatever the thread count. `--threads` defaults to the core count; on the Mac below, 8 threads took 5.1 s per DiT step against 3.8 s with all 16.
+
+## Python
+
+```python
+from ai_edge_litert.interpreter import Interpreter
```
smilingday: can we use
CompiledModel API?

We want to avoid use interpreter API, which is kept mostly for backforward compatability perpective.

smilingday (commented):
Thanks for doing this.

models/bonsai/bonsai_image_4b/README.md
john-rocky: Thanks for the pointer, and sorry: you had asked for the CompiledModel API over the Interpreter before, and I still sent this page with an Interpreter block, which was my slip. Both places now use the CompiledModel API: the block on this page and `python/generate.py`, which the page says makes the same calls, so the PR carries a script change as well as the page now (commit 38b150c3). The seed-0 image from the ported script is byte-identical to the Interpreter version's. The page's times were re-measured with it: the stage times moved by 0.1–0.2 s, and the whole image reads 20 s instead of 19 s, inside the 19.6–20.6 s the Interpreter version gives on the same Mac today. The copy of `generate.py` on the litert-community model card is updated to match, and the same commit repoints the source-checkpoint link at `prism-ml/bonsai-image-ternary-4B-unpacked`, the repo the conversion scripts load, since the old name no longer resolves on the Hub. Two other recipes of mine in the repo still use the Interpreter API (the KittenTTS streaming sample and a SAM2 video probe script); I'll move those to the CompiledModel API in a separate PR. Whatever you'd shape differently in the new block, I'll change. Thanks again.

PR #328, opened 2026-09-22
community/models_page: the litert-community models page and its checks

john-rocky:
A page that lists the litert-community org's models with their source model, the leaderboard's benchmarks and the recipe under `models/`, where these exist. It reads three sources when it loads: the org's model list on Hugging Face, and two files in this repository, the leaderboard's `board.json` and `models/README.md`. This adds the page's source under `community/models_page/` so it can be reviewed and changed in this repository: `space/` is the page, four files built as a static Hugging Face Space; `test/` its checks (node tests, a fixture server that runs the page with each source unreachable, slow, empty, malformed or hostile, and a check of the live sources); the README says how to run it locally. Nothing else changes. Once merged, GitHub Pages serves this directory at `community/models_page/space/`, the way it serves the leaderboard; it cannot be merged and kept off that URL, so this stays a Draft until the page should be public. `community/` is a new top-level directory and a first proposal, not a decision; next to the leaderboard under `benchmark/` may fit better, and whichever place fits, I will move it.

Thanks!

cc @smilingday


smilingday:
Our PM can give UI suggestions

smilingday (approved):

smilingday:
Let's get this in first, and we can iterate this page together.

PR #329, opened 2026-09-22
litert-compiled-model-migration: align API names with the current Kotlin sources

john-rocky:
Thanks for pointing me at this skill. Its frame — the upfront interview, the like-for-like first pass, the two-stage gate — is the right order for a migration, so the structure stays as it is; the two commits here change only facts and API names.

I checked each claim against the LiteRT Kotlin sources at v2.2.0, the 2.2.0 AARs and the Google Maven index, the C++ SDK zip of the v2.2.0 release and its docs page, the TensorFlow Lite 2.16.1 AARs, ai-edge-quantizer 0.8.0, and the NDK page on NNAPI. The corrections fall into four groups:
- Kotlin API: `run()` takes `List<TensorBuffer>`; there is no `runAsync`, no `createFromAhwb` and no ByteBuffer input (buffers come from the model and are filled with `writeFloat` / `writeInt8`).
- Interpreter side: `org.tensorflow.lite.nnapi.NnApiDelegate` (deprecated since Android 15, no Java/Kotlin entry point in 2.x) and `FlexDelegate` (no 2.x artifact ships it, so a Select-TF-ops model is re-exported).
- Build and NPU: the `litert:2.2.0` line and minSdk 24, the documented C++ SDK for CMake, the environment the NPU provider sets up itself, the compiler cache, and the int8 recipe with calibration.
- Procedure: the Stage 2 test is an instrumented test (`connectedDebugAndroidTest`), plus the README's real URL and the skills index row.

Three statements from my first commit were wrong and are fixed here: both `Environment.create` overloads exist, the custom-op header is `litert/cc/litert_custom_op_kernel.h`, and `run()` blocks on the CPU while it can return early on the GPU.

I also measured the effect on a small legacy classifier (Interpreter, GPU delegate, Support library) with two coding agents given only the skill text: with the original text, the weaker agent wrote `runAsync` / `createFromBuffer` code (10 compile errors against the 2.2.0 classes) and the stronger one kept `2.1.6` and skipped Phase 2 because the features it describes do not exist; with this text, both produced code that compiles, with the NPU environment, the serial dispatcher and buffer reuse in place.

Two things I could not settle from outside: the `DeepCoderLite` / `DeepInvestigatorLite` note in section 0.A (names of one agent harness; left for you to decide), and a device run — the test template compiles with kotlinc against the 2.2.0 classes, but nothing ran on a phone.

Marking this ready for review; if you would rather keep any original wording or fold this into a larger update, I will change it.

cc @smilingday


smilingday (approved):
Thanks for doing this!

PR #335, opened 2026-09-24
Add advanced DDP benchmarking Colab using the device-run CLI

akshat14195:
ddp_benchmark.ipynb covers the simple case through `litert benchmark --ddp`. This notebook shows the layer underneath: calling `gcloud alpha device-run sessions submit android-executable` directly, so the caller picks the binary, the flags, and the files pushed to the device.

It benchmarks Gemma 4 E2B with litert_lm_advanced_main, taking the binary and its shared libraries from the public release bucket at gs://litert/binaries/latest/android_arm64/litert_lm/ and the model from Hugging Face, and demonstrates both session shapes:

- Direct execution, where the binary is the session executable and --executable-env-vars supplies the LD_LIBRARY_PATH that GPU runs need in order to dlopen the OpenCL delegate. One session, one configuration.
- A wrapper script as the session executable, running CPU and GPU per device in a single session, with a discarded warm-up pass and a cooldown between configurations.

It then pulls logcat.txt back from the session bucket, since an android-executable job writes process output to logcat rather than to a separate log file, parses prefill and decode tokens/s, time to first token and peak memory, verifies the OpenCL delegate actually loaded, and renders a comparison table and charts.

Validated on DDP hardware across Galaxy S25 Ultra, SM-S947U1 and Pixel 9 Pro.

The README is updated to cover both notebooks and which one to reach for.


benchmark/developer_device_platform/README.md
```diff
@@ -1,22 +1,32 @@
 # Benchmark using Real Devices in Developer Device Platform
 
-[`ddp_benchmark.ipynb`](ddp_benchmark.ipynb) benchmarks [LiteRT](https://github.com/google-ai-edge/litert) (`.tflite`) and
+Two notebooks benchmark [LiteRT](https://github.com/google-ai-edge/litert) (`.tflite`) and
```
xuchangxxzz: nit: The wording seems weird with "Two notebooks benchmark". How about "Notebooks to benchmark LiteRT..."

benchmark/developer_device_platform/README.md
```diff
@@ -1,22 +1,32 @@
...
 [LiteRT-LM](https://github.com/google-ai-edge/LiteRT-LM) (`.litertlm`) models on real phones in
 [Developer Device Platform](https://docs.cloud.google.com/developer-device-platform/overview) (DDP), Google Cloud's managed
-device lab, from a Colab runtime. The notebook drives DDP through the [LiteRT CLI](https://github.com/google-ai-edge/LiteRT-CLI):
-`litert benchmark --ddp` uploads the model, reserves a device, runs the prebuilt `benchmark_model` (LiteRT) or the LiteRT-LM
-benchmark binary on it, and prints the results. Nothing is installed on your machine, and no device is attached to it.
+device lab, from a Colab runtime. Nothing is installed on your machine, and no device is attached to it.
+
+| Notebook | Use it when |
+| :--- | :--- |
+| [`ddp_benchmark.ipynb`](ddp_benchmark.ipynb) | You want a number for one model on one device. One command per run. |
```
xuchangxxzz: I think we should change the name of those two files. You could change the wording a bit, but we should mention 

- litert_cli_benchmark: powered by the DDP API that is easy to benchmark litert runtimes, native support.
- ddp_cli_benchmark: using DDP cli directly with litert runtime as an example, easily extending across several devices and switching over to other benchmark binaries for other runtimes as well

benchmark/developer_device_platform/README.md
```diff
@@ -29,6 +39,28 @@ Colab auth, so a project id is the only input.
 Both models come from [Hugging Face](https://huggingface.co/litert-community) and need no conversion. To benchmark your own
 model, point `litert benchmark` at a local `.tflite` or `.litertlm` file instead.
 
+## `ddp_benchmark_advanced.ipynb` — the `device-run` CLI directly
+
+Skips the LiteRT CLI and calls `gcloud alpha device-run sessions submit android-executable` itself, so you choose the binary,
```
xuchangxxzz: I would not say skip the LiteRT CLI. We need to tell the user more about why we want to skip the LiteRT CLI, and why they wanna read more using the DDP CLI directly. Here, I wanna cover 3 things:

1) using LiteRT runtime as an example, easily extending to other runtimes
2) more lower-level control, where we got the logcat and multiple files generated by the binary directly
3) across to multiple device fleets

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
```diff
@@ -0,0 +1,709 @@
...
+   "source": [
+    "# Advanced: benchmarking on DDP with the `device-run` CLI\n",
+    "\n",
+    "> Looking for the simple path? [`ddp_benchmark.ipynb`](ddp_benchmark.ipynb) benchmarks a single model on a single device with one `litert benchmark --ddp` command. **This** notebook drops down to the underlying `gcloud alpha device-run` CLI, which is what you need for multi-device sweeps, custom binaries, and full control over benchmark flags.\n",
+    "\n",
+    "This notebook runs [`litert_lm_advanced_main`](https://github.com/google-ai-edge/LiteRT-LM/blob/main/runtime/engine/litert_lm_advanced_main.cc)\n",
+    "against **Gemma 4 E2B** on physical Android phones, using\n",
+    "**Developer Device Platform (DDP)** to allocate the hardware.\n",
+    "\n",
+    "It demonstrates **two ways to call DDP**:\n",
```
xuchangxxzz: the chart below formatted weirdly

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
```diff
@@ -0,0 +1,709 @@
...
+    "    \"libLiteRtOpenClAccelerator.so\",\n",
+    "    \"libLiteRtTopKOpenClSampler.so\",\n",
+    "]\n",
+    "LIB_PUSH = [f\"{LITERT_LM}/{lib}={WORK}/{lib}\" for lib in DEVICE_LIBS]\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "## \ud83d\udcf1 3. Finding device IDs\n",
```
xuchangxxzz: I think we should make this part further simple and attach a link https://docs.cloud.google.com/developer-device-platform/device-catalog for user to learn more

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
```diff
@@ -0,0 +1,709 @@
...
+   "metadata": {
+    "id": "32Qpb2GJkqMO"
+   },
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "# @title Run a single benchmark directly\n",
+    "# @markdown One device, one backend, no shell script.\n",
+    "\n",
+    "DIRECT_DEVICE  = \"pa3q-35\" # @param {type:\"string\"}\n",
```
xuchangxxzz: In our example, we should at least pick 2-3 devices to show how easy to extend devices

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
```diff
@@ -0,0 +1,709 @@
...
+    "  --other-files-to-push=\"{','.join(push)}\" \\\n",
+    "  --executable-timeout=30m 2>&1\n",
+    "\n",
+    "print(\"\\n\".join(direct_output))\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "source": [
+    "# @title Adds some methods to parse the metrics\n",
```
xuchangxxzz: could we fold this code block by default?

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
```diff
@@ -0,0 +1,709 @@
...
+    "  --other-files-to-push=\"...=/data/local/tmp/model.litertlm\"\n",
+    "```\n",
+    "\n",
+    "In this approach, three specific flags orchestrate the entire execution:\n",
+    "\n",
+    "* **`--executable`**: The compiled C++ binary itself. No wrapper shell script is used.\n",
+    "* **`--executable-args`**: A *comma-separated* list of arguments passed directly to the binary. *(Note: If an argument's value naturally contains a comma, you must escape it using the `^SEP^` syntax).*\n",
+    "* **`--executable-env-vars`**: Because there is no shell environment to run `export LD_LIBRARY_PATH`, we inject environment variables directly through the API. This is crucial for GPU runs, as the binary needs to know where to dynamically load the OpenCL delegate (`.so`) we pushed to the device.\n",
+    "\n",
+    "### \ud83d\udcdd How Output is Captured\n",
```
xuchangxxzz: I think this is misleading? I thought the binary will also generate outputs and we kind of extracting metrics from binary output directly, no?

benchmark/developer_device_platform/README.md
akshat14195: Done, thanks.

benchmark/developer_device_platform/README.md
akshat14195: Done. The notebooks are now `litert_cli_benchmark.ipynb` and `ddp_cli_benchmark.ipynb`, and the README uses your descriptions. Both Colab badges point at the new names. The only change to the existing notebook is its badge URL.

benchmark/developer_device_platform/README.md
akshat14195: Agreed. The README and the notebook intro now give three reasons to use the device-run CLI directly: other runtimes such as LiteRT-LM, lower-level control (logcat plus any files the binary writes), and running across device fleets.

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
akshat14195: Replaced the table with a bullet list.

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
akshat14195: Done. It's now one paragraph that links the [device catalog](https://docs.cloud.google.com/developer-device-platform/device-catalog), plus the `devices list` command.

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
akshat14195: Done. Part 1 now takes a device list (three by default) and submits them in one session, so the results table has one row per device.

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
akshat14195: Done. The helper cell is collapsed by default.

benchmark/developer_device_platform/ddp_benchmark_advanced.ipynb
akshat14195: Good call. The binary now writes its metrics with `--metric_proto_file_path`, the file is pulled back with `--paths-to-pull`, and the notebook decodes that file instead of parsing logcat. Logcat is only used to confirm the GPU run actually used OpenCL.

smilingday (approved):
Looks great!

